Issue #1, opened 2015-05-13
API: paradigm should be dataframes


GaelVaroquaux:
Looking at the code in experimental_paradigm.py:

- paradigms objects are really data containers (and fairly simple ones)

- the code in the module is for I/O of these to CSV files, and for checking their consistency

- there are 2 types of paradigms, event and block

- paradigms are are defined by a small number of named lists that have the same number of entries:
     - event: cond_id, onset, amplitude 
     - block: cond_id, onset, duration and amplitude

All this makes me think that paradigm objects could be replaced by pandas dataframe with the corresponding entries, and a few functions, in particular something like a '_check_paradigm' for our own internal usage that would take a pandas dataframe and validated it as a block or event paradigm and if it is not valid raise an understandable error. Pandas would provide the reading and writing CSVs.

bthirion:
Solved. closing.

PR #3, opened 2015-05-21
Dataframe

bthirion:
Addresses #1 

Took this opportunity to clean up test a little bit.


continuous_integration/install.sh
```diff
@@ -106,4 +106,6 @@ fi
 
 pip install nilearn
 
+pip install pandas
```
lesteve: I don't think you need this, right? Pandas should be either installed via conda or apt-get in the neurodebian case.

examples/plot_localizer_analysis.py
```diff
@@ -54,8 +52,11 @@
 # Design matrix
 ########################################
 
-paradigm = load_paradigm_from_csv_file(paradigm_file)['0']
-
+paradigm = DataFrame.from_csv(paradigm_file)
+onset = [float(x[0].split(' ')[2]) for x in paradigm.iterrows()]
+name = [x[0].split(' ')[1] for x in paradigm.iterrows()]
+n_conditions = len(np.unique(name))
+paradigm = DataFrame({'onset': onset, 'name': name})
```
lesteve: You can greatly simplify this. I don't know what the first column (with only zeros) is so you can chose a better name if you want:

```python
df = DataFrame.from_csv(paradigm_file, sep=' ', header=None, index_col=None)
df.columns = ['no_clue_what_so_ever', 'name', 'onset']
n_conditions = len(df.name.unique())
```

bthirion: Thx for the tip. I'm essentially new at pandas. 

----- Mail original -----

continuous_integration/install.sh
bthirion: For some reason, this does not works. Would you mind taking a look at 
https://travis-ci.org/nistats/nistats/builds/66196889 

AFAIK, since we do "conda update --yes conda" install.sh, pandas install is supposed to work... 
?? 

----- Mail original -----

lesteve: You probably need to define PANDAS_VERSION in .yaml.

bthirion:
Took me some time to get it right, but it works now.

.travis.yml
```diff
@@ -13,21 +13,24 @@ env:
...
       SCIKIT_LEARN_VERSION="0.14.1"
+      PANDAS_VERSION="0.13.0"
     - DISTRIB="neurodebian" PYTHON_VERSION="2.7"
     # Trying to get as close to the minimum required versions while
     # still having the package version available through conda
     - DISTRIB="conda" PYTHON_VERSION="2.6"
       NUMPY_VERSION="1.6.2" SCIPY_VERSION="0.11.0"
       SCIKIT_LEARN_VERSION="0.12.1" MATPLOTLIB_VERSION="1.1.1"
-      NIBABEL_VERSION="1.1.0"
+      NIBABEL_VERSION="1.1.0" PANDAS_VERSION="0.9.0"
```
lesteve: In my local tests, it seems you can get pandas 0.12 with numpy 1.6.2 through conda. Any reason why to use 0.9 here? Just asking because I seem to remember the minimum pandas required version was 0.14 previously.

README.rst
```diff
@@ -25,12 +25,13 @@ Dependencies
...
 The required dependencies to use the software are:
 
-* Python >= 2.6,
+* Python >= 2.6
 * setuptools
 * Numpy >= 1.6.1
 * SciPy >= 0.9
 * Nibabel >= 1.1.0
 * Nilearn >= 0.1.3
+* Pandas >= 0.14
```
lesteve: Depending on which version of pandas ends up being required this line should be eventually changed accordingly.

nistats/tests/test_dmtx.py
```diff
@@ -293,7 +302,7 @@ def test_design_matrix10():
     X, names = design_matrix_light(frame_times, paradigm, hrf_model=hrf_model,
                          drift_model='polynomial', drift_order=3,
                          fir_delays=range(1, 5))
-    onset = paradigm.onset[paradigm.con_id == 'c0'].astype(np.int)
+    onset = paradigm['onset'][paradigm['name'] == 'c0'].astype(np.int)
```
lesteve: Minor remark: you can use the dot notation to access columns in pandas (as long as your column name is a valid python identifier IIRC). So I reckon this should work:

```python
paradigm.onset[paradigm.name == 'c0']
```

nistats/version.py
```diff
@@ -44,7 +44,12 @@
         'install_info': _NISTATS_INSTALL_MSG}),
     ('nibabel', {
         'min_version': '1.1.0',
-        'required_at_installation': False}))
+        'required_at_installation': False}),
+    ('pandas', {
+        'min_version': '0.9.0',
```
lesteve: Just using this as a reminder here to set the pandas version consistently here too.

bthirion:
@lesteve I think I have addressed your comments.

nistats/version.py
lesteve: I guess 0.12 would be in order here. I'll fix that after this PR is merged.

lesteve:
Thanks merging.

PR #4, opened 2015-05-27
MRG: Better naming

bthirion:
Started to use more conventional, less jargon naming for many variables

### DIFF
--- a/nistats/hemodynamic_models.py
+++ b/nistats/hemodynamic_models.py
@@ -61,10 +61,14 @@ def glover_hrf(tr, oversampling=16, time_length=32., onset=0.):
 
     Parameters
     ----------
-    tr: float, scan repeat time, in seconds
-    oversampling: int, temporal oversampling factor, optional
-    time_length: float, hrf kernel length, in seconds
-    onset: float, onset of the response
+    tr: float
+        scan repeat time, in seconds
+    oversampling: int
+        temporal oversampling factor, optional
+    time_length: float
+        hrf kernel length, in seconds
+    onset: float
+        onset of the response
 
     Returns
     -------
@@ -81,10 +85,14 @@ def spm_time_derivative(tr, oversampling=16, time_length=32., onset=0.):
 
     Parameters
     ----------
-    tr: float, scan repeat time, in seconds
-    oversampling: int, temporal oversampling factor, optional
-    time_length: float, hrf kernel length, in seconds
-    onset: float, onset of the response
+    tr: float
+        scan repeat time, in seconds
+    oversampling: int
+        temporal oversampling factor, optional
+    time_length: float
+        hrf kernel length, in seconds
+    onset: float
+        onset of the response
 
     Returns
     -------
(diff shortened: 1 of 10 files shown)


nistats/hemodynamic_models.py
```diff
@@ -148,45 +148,45 @@ def _sample_condition(exp_condition, frametimes, oversampling=16,
...
+        events that start before frame_times[0] + min_onset are not considered
 
     Returns
     -------
     regressor: array of shape(n)
         possibly oversampled event regressor
-    hr_frametimes : array of shape(n)
-        frametimes corresponding to regressor
+    hr_frame_times : array of shape(n)
+        frame_times corresponding to regressor
```
lesteve: Should this one be left as it was or maybe spelled in two words "frame times"?

nistats/tests/test_hemodynamic_models.py
```diff
@@ -215,17 +215,17 @@ def test_design_warnings():
...
-                     frametimes)
+                     frame_times)
     condition = ([-25, -25, 36.5], [0, 0, 0], [1, 1, 1])
     with warnings.catch_warnings(record=True):
         warnings.simplefilter("always")
         assert_warns(UserWarning, compute_regressor, condition, hrf_model, 
-                     frametimes)
+                     frame_times)
 
 if __name__ == "__main__":
```
lesteve: just curious, is that here only to be able to do `python this_test.py` rather than `nosetests this_test.py` ? Not sure I see a good reason not to always use the latter.

nistats/hemodynamic_models.py
```diff
@@ -197,19 +197,19 @@ def _sample_condition(exp_condition, frametimes, oversampling=16,
...
     regressor = np.cumsum(regressor)
 
-    return regressor, hr_frametimes
+    return regressor, hr_frame_times
 
 
-def _resample_regressor(hr_regressor, hr_frametimes, frametimes, kind='linear'):
-    """ this function samples the regressors at frametimes
+def _resample_regressor(hr_regressor, hr_frame_times, frame_times, kind='linear'):
+    """ this function samples the regressors at frame_times
```
lesteve: Maybe use \`\`frame_times\`\` if you meant to use the variable name or "frame times"/"frametimes" if you didn't.

nistats/hemodynamic_models.py
```diff
@@ -343,16 +343,16 @@ def compute_regressor(exp_condition, hrf_model, frametimes, con_id='cond',
     'canonical_derivative': the Glover hrf + time derivative (2 regressors)
     'fir': finite impulse response basis, a set of delayed dirac models
            with arbitrary length. This one currently assumes regularly spaced
-           frametimes (i.e. fixed time of repetition).
+           frame_times (i.e. fixed time of repetition).
```
lesteve: Same kind of comments as previous docstrings: do we want to use some words here or the variable name.

bthirion:
Thx for the review !
 I think that the latest commits have addressed your comments + additional cosmits (tests, docstrings)

lesteve:
Looks good to me, merging.

PR #9, opened 2015-07-15
Spm example

bthirion:
Added dowloaders + working examples for a couple of datasets:
* SPM auditory
* SPM multimodal faces


bthirion:
I'll merge that one tomorrow if nobody objects...

Issue #14, opened 2015-07-22
Better test datasets fetching


bthirion:
Re-use Nilearn infrastructure.

PR #17, opened 2015-07-29
WIP: Test datasets

bthirion:
First attempt to address #14 : better tests of datasets module.

The big inconvenient is that I had to introduce the mock setup in the dataset functions, which is really unclean.
The ideal thing is to reuse the _fetch_files function of nilearn, but this may require a deeper refactoring of the code.


bthirion:
Idem solved by #74. Cloding.

Issue #5, opened 2015-06-12
Add more examples of GLM fit


bthirion:
Based on public datasets. Try to vary the setting (block, event, phase-encoding, multi-session, possibly multi-subject).

bthirion:
partly OK now. See also #12

PR #18, opened 2015-08-17
MRG: Adding an example with FIAC dataset

bthirion:
Addresses #5

### DIFF
--- a/examples/plot_fiac_analysis.py
+++ b/examples/plot_fiac_analysis.py
@@ -1,0 +1,99 @@
+_doc = """
+GLM fitting in fMRI
+===================
+
+Full step-by-step example of fitting a GLM to experimental data and visualizing
+the results. This is done on two runs of one subject of the FIAC dataset.
+For details on the data, please see:
+ 
+Dehaene-Lambertz G, Dehaene S, Anton JL, Campagne A, Ciuciu P, Dehaene
+G, Denghien I, Jobert A, LeBihan D, Sigman M, Pallier C, Poline
+JB. Functional segregation of cortical language areas by sentence
+repetition. Hum Brain Mapp. 2006: 27:360--371.
+http://www.pubmedcentral.nih.gov/articlerender.fcgi?artid=2653076#R11
+
+
+More specifically:
+
+1. A sequence of fMRI volumes are loaded
+2. A design matrix describing all the effects related to the data is computed
+3. a mask of the useful brain volume is computed
+4. A GLM is applied to the dataset (effect/covariance,
+   then contrast estimation)
+"""
+
+print _doc
+
+from os import mkdir, path, getcwd
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+from nilearn.plotting import plot_stat_map
+import nibabel as nib
+
+from nistats.glm import FMRILinearModel
+from nistats.design_matrix import make_design_matrix
+from nistats import datasets
+
+
+# write directory
+write_dir = path.join(getcwd(), 'results')
+if not path.exists(write_dir):
+    mkdir(write_dir)
+
+# Data and analysis parameters
+data = datasets.fetch_fiac_first_level()
+fmri_files = [data['func1'], data['func2']]
+design_files = [data['dmtx_ses1'], data['dmtx_ses2']]
+
+# Load all the data into a common GLM
+multi_session_model = FMRILinearModel(fmri_files, design_files, data['mask'])
+
+# GLM fitting
+multi_session_model.fit(do_scaling=True, model='ar1')
+
+
+def make_fiac_contrasts(n_columns):
+    """ Specify some contrasts for the FIAC experiment"""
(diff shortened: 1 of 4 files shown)


bthirion:
This one should be mergeable now.

examples/plot_fiac_analysis.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # make a snapshot of the contrast activation
+    if contrast_id == 'Effects_of_interest':
+        vmax = max(- z_map.get_data().min(), z_map.get_data().max())
+        vmin = - vmax
+        plot_stat_map(z_map, bg_img=mean_img, threshold=2.5, title=contrast_id)
+        plt.savefig(path.join(write_dir, '%s_z_map.png' % contrast_id))
+
+print('All the  results were witten in %s' % write_dir)
+plt.show()
```
lesteve: You don't seem to do any plotting in this example. If that's on purpose, you can remove this line and the matplotlib import.

examples/plot_fiac_analysis.py
```diff
@@ -0,0 +1,99 @@
+_doc = """
```
lesteve: Why use a module docstring and then use `print(__doc___)` below?

examples/plot_fiac_analysis.py
```diff
@@ -0,0 +1,99 @@
...
+More specifically:
+
+1. A sequence of fMRI volumes are loaded
+2. A design matrix describing all the effects related to the data is computed
+3. a mask of the useful brain volume is computed
+4. A GLM is applied to the dataset (effect/covariance,
+   then contrast estimation)
+"""
+
+print _doc
```
lesteve: Use parentheses for Python3 compatibility.

examples/plot_fiac_analysis.py
lesteve: My bad I had a lingering matplotlibrc setting the backend to Agg.

nistats/datasets.py
```diff
@@ -237,8 +237,68 @@ def _glob_spm_multimodal_fmri_data():
...
+        """glob data from subject_dir."""
+        _subject_data = {}
+        subject_dir = os.path.join(data_dir, 'nipy-data-0.2/data/fiac/fiac0')
+        for session in [1, 2]:
+            # glob func data for session s + 1
+            session_func = os.path.join(subject_dir, 'run%i.nii.gz' % session)
+            if not os.path.isfile(session_func):
+                print('Missing functional scan for session %i.' % session)
+                return None
+            else:
```
lesteve: you don't really need the `else` and can dedent the following line one level since you have an early return in the if clause. You have a few similar patterns below.

nistats/datasets.py
```diff
@@ -237,8 +237,68 @@ def _glob_spm_multimodal_fmri_data():
...
+    print('Data absent, downloading...')
+    url = 'http://nipy.sourceforge.net/data-packages/nipy-data-0.2.tar.gz'
+
+    archive_path = os.path.join(data_dir, os.path.basename(url))
+    _fetch_file(url, data_dir)
+    try:
+        _uncompress_file(archive_path)
+    except:
+        print('Archive corrupted, trying to download it again.')
+        return fetch_fiac_first_level(data_dir=data_dir)
```
lesteve: Just curious, why do you need both fetch_fiac_first_level and _glob_fiac_data, is that to have a quick and dirty retry mechanism?

examples/plot_fiac_analysis.py
```diff
@@ -0,0 +1,99 @@
...
+    z_map, = multi_session_model.contrast(
+        [contrast_val] * 2, con_id=contrast_id, output_z=True)
+    nib.save(z_map, z_image_path)
+
+    # make a snapshot of the contrast activation
+    if contrast_id == 'Effects_of_interest':
+        vmax = max(- z_map.get_data().min(), z_map.get_data().max())
+        vmin = - vmax
+        plot_stat_map(z_map, bg_img=mean_img, threshold=2.5, title=contrast_id)
+        plt.savefig(path.join(write_dir, '%s_z_map.png' % contrast_id))
```
lesteve: Rather than using plt.savefig, I guess the "nilearn way" would be to do:

```python

display = plot_stat_map(z_map, bg_img=mean_img, threshold=2.5, title=contrast_id)
display.savefig(path.join(write_dir, '%s_z_map.png' % contrast_id))
```

nistats/datasets.py
bthirion: Yes, I think that this is the main reason. Note that I copy-pasted that part from Pypreprocess without thinking deeply of it. When Nilearn is released, I or whoever wants will refactor this code to make it cleaner.

examples/plot_fiac_analysis.py
bthirion: Sure.

bthirion:
Should be good now. Thx for the comments.

Issue #19, opened 2015-08-18
Some functions in utils.py are never used


bthirion:
open4csv

bthirion:
Should have been closed earlier.

PR #20, opened 2015-08-18
MRG: Delete unused functions

bthirion:
Addresses #19

### DIFF
--- a/nistats/utils.py
+++ b/nistats/utils.py
@@ -107,7 +107,7 @@ def multiple_mahalanobis(effect, covariance):
 
     # compute the inverse of the covariances
     Kt = multiple_fast_inv(Kt)
-    
+
     # derive the squared Mahalanobis distances
     sqd = np.sum(np.sum(Xt[:, :, np.newaxis] * Xt[:, np.newaxis] * Kt, 1), 1)
     return sqd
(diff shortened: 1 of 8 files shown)


nistats/utils.py
```diff
@@ -9,15 +9,12 @@
...
-TINY = 1e-15
-
 
 def z_score(pvalue):
     """ Return the z-score corresponding to a given p-value.
     """
-    pvalue = np.minimum(np.maximum(pvalue, TINY), 1. - TINY)
-    z = norm.isf(pvalue)
-    return z
+    pvalue = np.minimum(np.maximum(pvalue, 1.e-300), 1. - 1.e-16)
```
lesteve: You replaced `1 - 1e-15` by `1 - 1e-16`. Was that on purpose?

lesteve: Ah I also missed the fact that you used 1e-300 for TINY inside the max, not sure why.

bthirion: The purpose of TINY is to avoid numerical underlows. It turns out that the 1.e-15 used previously was not 'small' enough and created unpleasant numerical artifacts. I now use smaller values while still avoiding numerical instabilities.

lesteve: I see. Is there an easy way to add a test for these numerical issues?

bthirion: It is there...

lesteve: My only worry is that the test is a bit loose given that you were able to change the magic numbers without making it fail. Having said that I am no expert in numerical instability.

PR #29, opened 2015-09-15
MRG: clean and removed unused code

bthirion:
* full_rank was duplicated
* matrix_rank us currently available in  np.linalg


MartinPerez:
All the code seems okay to me.

nistats/utils.py
```diff
@@ -113,109 +114,36 @@ def multiple_mahalanobis(effect, covariance):
...
-
-    Parameters
-    ----------
-    X : array-like
-        2D array which may not be of full rank.
-    r : None or int
-        Known rank of `X`.  r=None results in standard matrix rank calculation.
-        We do not check `r` is really the rank of X; it is to speed up
-        calculations when the rank is already known.
+    cmax : float, optional (default:1.e-15),
```
lesteve: is the default supposed to be 1e+15 or 1e-15. Doc says 1e-15 but function signature says 1e+15.

nistats/utils.py
```diff
@@ -113,109 +114,36 @@ def multiple_mahalanobis(effect, covariance):
...
+        tolerance for condition number
 
     Returns
     -------
-    fX : array
-        Full-rank matrix with column span matching that of `X`
+    X : array of shape(nrows, ncols)
```
lesteve: missing space between shape and the parenthesis

nistats/utils.py
```diff
@@ -113,109 +114,36 @@ def multiple_mahalanobis(effect, covariance):
...
-    if r is None:
-        r = matrix_rank(X)
-    V, D, U = spl.svd(X, full_matrices=0)
-    order = np.argsort(D)
-    order = order[::-1]
-    value = []
-    for i in range(r):
-        value.append(V[:, order[i]])
-    return np.asarray(value).T.astype(np.float64)
+    U, s, V = spl.svd(X, 0)
```
lesteve: Use full_matrices=False which is clearer than a magical 0.

nistats/utils.py
lesteve: Reading the code a bit more, it looks like the doc is wrong and it should say 1e+15.

nistats/utils.py
```diff
@@ -113,109 +114,36 @@ def multiple_mahalanobis(effect, covariance):
...
-    M = np.asarray(M)
-    if M.ndim > 2:
-        raise TypeError('array should have 2 or fewer dimensions')
-    if M.ndim < 2:
-        return int(not np.all(M == 0))
-    S = spl.svd(M, compute_uv=False)
-    if tol is None:
-        tol = S.max() * np.finfo(S.dtype).eps * max(M.shape)
-    return np.sum(S > tol)
+    X : array of shape(nrows, ncols)
```
lesteve: missing space between shape and parenthesis

Issue #36, opened 2015-11-18
[enhancement] Solving issues with current hemodynamic models


bthirion:
I see at least two issues:
* the naming is inconsistent e.g.
* the values for time derivatives depend on the oversampling factor, which is rather weird.

MartinPerez:
@bthirion shall we close this issue?

PR #37, opened 2015-11-21
MRG: hrf refactor

bthirion:
This is an update of hrf module. Fixes a numerical issue, improves docstrings, adds  an example.
Addresses #36 and #34. 
The most debattable change is the removal of the name 'canonical' for the 'glover' hrf, because this name is unclear (in spm, the canonical hrf is not that one). This will break the compatibility with nipy.
What do people think about that ?

### DIFF
--- a/nistats/design_matrix.py
+++ b/nistats/design_matrix.py
@@ -16,6 +16,11 @@ Design matrices contain three different types of regressors:
    A hemodynamic model is one of:
    'spm' : linear filter used in the SPM software
    'glover' : linear filter estimated by G.Glover
+   'spm + derivative', 'glover + derivative': the same linear models,
+       plus their time derivative (2 regressors per condition)
+   'spm + derivative + dispersion', 'glover + derivative + dispersion':
+       idem plus the derivative wrt the dispersion parameter of the hrf
+       (3 regressors per condition)
    'fir' : finite impulse response model, generic linear filter
 
 2. User-specified regressors, that represent information available on
@@ -168,7 +173,8 @@ def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
         for these to be valid paradigm descriptors
 
     hrf_model : {'spm', 'spm + derivative', 'spm + derivative + dispersion',
-        'glover', 'glover + derivative', 'fir'}
+        'glover', 'glover + derivative', 'glover + derivative + dispersion',
+        'fir'}
         String that specifies the hemodynamic response function
 
     frame_times : array of shape (n_scans,)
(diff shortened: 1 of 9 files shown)


MartinPerez:
Why not keep the canonical (the one implemented in spm) and then also add the glover in the documentation and examples as an option? Someone like working with fmri experiments would like to see a link to some paper or something that justifies the deviation from the SPM standards that are hardwired in the community. I can imagine a reviewer asking for the SPM canonical results alongside the glover one.

bthirion:
I understand your point, but so far, what was called the 'canonical' hrf so far was the 'Glover' one !
This means that the term "canonical" is actually not so canonical...
So I prefer not to use it any more. Taking the spm hrf as a default feels right. 
I can change my mind if people believe that 'canonical' should be used instead (talking about the 'spm' hrf).
ping @agramfort ?

agramfort:
canonical for spm or for freesurfer does not mean the same thing.

I would avoid canonical. It's not specific.

MartinPerez:
all here seems fine to me

bthirion:
No further comment on that one ? Note that it breaks back-ward compatibility with nipy...

nistats/design_matrix.py
```diff
@@ -271,9 +276,9 @@ def make_design_matrix(
     paradigm : DataFrame instance, optional
         Description of the experimental paradigm.
 
-    hrf_model : string, optional,
-        Specifies the hemodynamic response function (HRF).
-        It can be 'canonical', 'canonical with derivative' or 'fir'
+    hrf_model : {'spm', 'spm + derivative', 'spm + derivative + dispersion',
+        'glover', 'glover + derivative', 'fir'}, optional,
```
MartinPerez: 'glover + derivative + dispersion' missing?

nistats/hemodynamic_models.py
```diff
@@ -327,15 +329,15 @@ def _regressor_names(con_name, hrf_model, fir_delays=None):
     names: list of strings,
         regressor names
     """
-    if hrf_model == 'canonical':
+    if hrf_model == 'glover':
         return [con_name]
-    elif hrf_model == "canonical with derivative":
+    elif hrf_model == "glover + derivative":
         return [con_name, con_name + "_derivative"]
```
MartinPerez: 'glover + derivative + dispersion' missing?

nistats/tests/test_dmtx.py
```diff
@@ -176,7 +176,7 @@ def test_design_matrix1b():
     tr = 1.0
     frame_times = np.linspace(0, 127 * tr, 128)
     paradigm = basic_paradigm()
-    hrf_model = 'Canonical'
+    hrf_model = 'Glover'
```
MartinPerez: 'Glover' with uppercaps instead of 'glover' like everywhere else.

MartinPerez:
Just one question, any reason the glover has no dispersion option?

bthirion:
I think I have addressed your comments. Thx.

MartinPerez:
@bthirion can we merge this? to avoid having inconsistent hrf naming in coming api changes and examples.

Issue #41, opened 2015-11-23
[enhancement] Integrate frame_times into design matrix


bthirion:
one idea I add also would be to use the index of a DataFrame rather than using the frame_times (don't like the name much btw)

In [4]: df = pd.DataFrame(dict(reg1=np.random.randn(10), reg2=np.random.randn(10)))

In [5]: df
Out[5]:
       reg1      reg2
0  0.435177 -1.034931
1 -0.059424  2.223872
2  0.070459 -0.838222
3 -0.777904 -0.690715
4 -0.782690 -0.045528
5  0.278254  0.388872
6  0.867945  0.242028
7 -2.214093  0.866546
8 -0.186639 -0.626605
9 -0.635547  1.209869

In [6]: tr = 2.4

In [7]: frame_times = np.arange(10) * 2.4

In [8]: df.index = frame_times

In [9]: df
Out[9]:
          reg1      reg2
0.0   0.435177 -1.034931
2.4  -0.059424  2.223872
4.8   0.070459 -0.838222
7.2  -0.777904 -0.690715
9.6  -0.782690 -0.045528
12.0  0.278254  0.388872
14.4  0.867945  0.242028
16.8 -2.214093  0.866546
19.2 -0.186639 -0.626605
21.6 -0.635547  1.209869

In [10]: matrix = df.values

In [11]: matrix
Out[11]:
array([[ 0.43517668, -1.03493107],
       [-0.05942358,  2.22387234],
       [ 0.07045882, -0.83822237],
       [-0.77790391, -0.69071475],
       [-0.78269037, -0.04552835],
       [ 0.27825373,  0.38887175],
       [ 0.8679447 ,  0.2420283 ],
       [-2.21409337,  0.8665459 ],
       [-0.18663918, -0.62660464],
       [-0.63554651,  1.20986888]])

Issue #40, opened 2015-11-22
use design matrix object in the GLM API


bthirion:
See comment in #39

PR #42, opened 2015-12-01
Design matrix glm api

bthirion:
Addresses issues #40 and #41 + remove a little bit of code in examples.
Overall, we rely more systematically on DataFrames to store information on design.

### DIFF
--- a/examples/plot_fiac_analysis.py
+++ b/examples/plot_fiac_analysis.py
@@ -29,6 +29,7 @@ from os import mkdir, path, getcwd
 
 import numpy as np
 import matplotlib.pyplot as plt
+import pandas as pd
 
 from nilearn.plotting import plot_stat_map
 from nilearn.image import mean_img
@@ -47,6 +48,8 @@ if not path.exists(write_dir):
 data = datasets.fetch_fiac_first_level()
 fmri_files = [data['func1'], data['func2']]
 design_files = [data['design_matrix1'], data['design_matrix2']]
+design_files = [pd.DataFrame(np.load(df)['X']) for df in design_files]
+
 
 # Load all the data into a common GLM
 multi_session_model = FirstLevelGLM(data['mask'], standardize=False,
@@ -77,7 +80,7 @@ def make_fiac_contrasts(n_columns):
     return contrast
 
 # compute fixed effects of the two runs and compute related images
-n_columns = np.load(design_files[0])['X'].shape[1]
+n_columns = design_files[0].shape[1]
 contrasts = make_fiac_contrasts(n_columns)
 
 print('Computing contrasts...')
--- a/examples/plot_fiac_analysis.py
+++ b/examples/plot_fiac_analysis.py
@@ -29,9 +29,8 @@ from os import mkdir, path, getcwd
 
 import numpy as np
 import pandas as pd
-import matplotlib.pyplot as plt
 
-from nilearn.plotting import plot_stat_map
+from nilearn import plotting
 from nilearn.image import mean_img
 import nibabel as nib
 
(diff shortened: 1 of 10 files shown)


agramfort:
cool :)

examples/plot_fiac_analysis.py
```diff
@@ -47,13 +48,15 @@
...
+design_matrices = [pd.DataFrame(np.load(df)['X']) for df in design_files]
+
 
 # Load all the data into a common GLM
 multi_session_model = FirstLevelGLM(data['mask'], standardize=False,
                                     noise_model='ar1')
 
 # GLM fitting
-multi_session_model.fit(fmri_files, design_files)
+multi_session_model.fit(fmri_files, design_matrices)
```
MartinPerez: Why we have to clarify that is a multi_session_model? it sounds to me like I have to do something different for multi_session and single_session, when in reality is just passing a list or single variables as parameters to glm_model.fit(imgs, design_matrices). Also in some examples we use epi_img, here fmri_files, etc, shouldnt we keep a name convention in the examples, it helps me a lot as a novice user.

examples/plot_localizer_analysis.py
```diff
@@ -47,46 +42,29 @@
...
-ax = plot_design_matrix(design_matrix)
-ax.set_position([.05, .25, .9, .65])
-ax.set_title('Design matrix')
-plt.savefig(path.join(write_dir, 'design_matrix.png'))
+### Perform a GLM analysis ########################################
 
-########################################
-# Perform a GLM analysis
-########################################
+fmri_glm = FirstLevelGLM().fit(epi_img, design_matrix)
```
MartinPerez: Same on my previous comment, I would ask for standard img name across examples

examples/plot_spm_auditory.py
```diff
@@ -48,25 +48,17 @@
...
+contrasts = dict([(column, contrast_matrix[i])
+                  for i, column in enumerate(design_matrix.columns)])
 # Use a  more interesting contrast
 contrasts = {'active-rest': contrasts['active'] - contrasts['rest']}
 
 # fit GLM
 print('\r\nFitting a GLM (this takes time) ..')
 fmri_glm = FirstLevelGLM(noise_model='ar1', standardize=False).fit(
-    [subject_data.func], matrix)
+    [subject_data.func], design_matrix)
```
MartinPerez: Same as previous comment for the standard name, maybe to simplify reading we should have a line for
img = subject_data.func then pass img?

examples/plot_spm_multimodal_faces.py
```diff
@@ -70,24 +70,20 @@
...
-
+contrast_matrix = np.eye(design_matrix.shape[1])
+contrasts = dict([(column, contrast_matrix[i])
+                  for i, column in enumerate(design_matrix.columns)])
 # more interesting contrasts
-contrasts['faces-scrambled'] = contrasts['faces'] - contrasts['scrambled']
-contrasts['scrambled-faces'] = -contrasts['faces-scrambled']
-contrasts['effects_of_interest'] = np.vstack((contrasts['faces'],
-                                              contrasts['scrambled']))
+interesting_contrasts = {
```
MartinPerez: It is a bit confusing for me reading the example, what is the point of separating contrasts and interesting contrasts and then only passing interesting contrasts to the model, also is kind of non standard to do that, normally contrast you define, contrast you pass. I dont understand why to change this code in the example.

nistats/design_matrix.py
```diff
@@ -4,10 +4,10 @@
 """
 This module implements fMRI Design Matrix creation.
 
-The DesignMatrix object is just a container that represents the design matrix.
+Design matrices a rerepsented by Pandas DataFrames
```
MartinPerez: rerepsented -> are represented

MartinPerez:
Except for a typo and some style questions, the code looks good to me. The examples are really a lot clearer to me with these changes.

bthirion:
Thanks. I'll look at that more in detail soonish.

examples/plot_spm_multimodal_faces.py
bthirion: The point is to avoid clutter by showing figures on 25 contrasts...
What we can do instead is to simplify the code. Please comment on my commit.

bthirion:
I think I have addressed most of your comments. Thx. More feedback welcome.

bthirion:
The examples are borken on master, so I have to merge that one.

Issue #21, opened 2015-08-19
Add CircleCI support


bthirion:
Watch the experience of nilearn/pypreprocess

souravsingh:
I would like to work on the issue.

bthirion:
That wuold be awesome, thanks. I believe that much of the configuration can be borrowed from Nielarn.

PR #46, opened 2015-12-15
[RFC]Created circle.yml for CircleCI

souravsingh:
Created a file for CircleCI testing

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,0 +1,35 @@
+dependencies:
+  cache_directories:
+    - "~/nistats_data"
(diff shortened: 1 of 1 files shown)


souravsingh:
This PR is related to https://github.com/nistats/nistats/issues/21

bthirion:
This sounds good, but the point is to feed the documentation, that remains currently empty.
For the moment, I would suggest to rely on sphync-gallery and simply run the examples (see Nilearn).
Best,

PR #35, opened 2015-11-17
[MRG] Statistical mapping

bthirion:
Added elementary utilities for thresholding. Addresses part of #28

PR #51, opened 2016-02-01
basic first level permutation tests

MartinPerez:
Here I provide the localizer example modified to perform basic contrasts first level permutation tests. I also provide an initial proposal for a permutation API. I am convinced this API is far from optimal, particularly I find cumbersome the way I am passing the design matrix objs to the permutation test. I would support #40, since I think having a design matrix object would allow me to pass it and access all relevant information from it as I do from the FirstLevelGLM object.

I am not sure on what would be the appropriate way to write tests for this functionality. In any case first eye tests seem nice. Following we can see the comparison between the original z score map for the contrast reading-visual for uncorrected p<0.1, then the permuted version for uncorrected p<0.1 and finally the permuted version for corrected p<0.5. The correction is very strong. 600 permutations were performed.

reading-visual_z_map

reading-visual_unc_pval

reading-visual_cor_pval

### DIFF
--- a/examples/plot_localizer_analysis_permutation.py
+++ b/examples/plot_localizer_analysis_permutation.py
@@ -1,0 +1,148 @@
+"""
+GLM fitting in fMRI
+===================
+
+Full step-by-step example of fitting a GLM to experimental data realizing
+permutation tests and visualizing the results.
+
+More specifically:
+
+1. A sequence of fMRI volumes are loaded
+2. A design matrix describing all the effects related to the data is computed
+3. a mask of the useful brain volume is computed
+4. A GLM is applied to the dataset (effect/covariance,
+   then contrast estimation)
+5. Permutation tests are performed to obtain uncorrected and corrected p values
+at each voxel for a contrast comparing only two conditions.
+
+"""
+
+print(__doc__)
+
+from os import mkdir, path
+
+import numpy as np
+import pandas as pd
+from nilearn import plotting
+
+from nistats.glm import FirstLevelGLM
+from nistats.design_matrix import make_design_matrix
+from nistats import datasets
+from nistats.permutation_tests import first_level_permutation_test
+from nistats.utils import z_score
+import nibabel as nib
+
+
+n_proc = 6
+
+### Data and analysis parameters #######################################
+
+# timing
+n_scans = 128
+tr = 2.4
+frame_times = np.linspace(0.5 * tr, (n_scans - .5) * tr, n_scans)
+
+# data
+data = datasets.fetch_localizer_first_level()
+paradigm_file = data.paradigm
+fmri_img = data.epi_img
+
+### Design matrix ########################################
+
+paradigm = pd.read_csv(paradigm_file, sep=' ', header=None, index_col=None)
+paradigm.columns = ['session', 'name', 'onset']
+design_matrix = make_design_matrix(
+    frame_times, paradigm, hrf_model='canonical with derivative',
+    drift_model="cosine", period_cut=128)
+
+### Perform a GLM analysis ########################################
+
+fmri_glm = FirstLevelGLM().fit(fmri_img, design_matrix)
+
+### Estimate contrasts #########################################
+
+# Specify the contrasts
+contrast_matrix = np.eye(design_matrix.shape[1])
+contrasts = dict([(column, contrast_matrix[i])
+                  for i, column in enumerate(design_matrix.columns)])
+
+contrasts["audio"] = contrasts["clicDaudio"] + contrasts["clicGaudio"] +\
+    contrasts["calculaudio"] + contrasts["phraseaudio"]
+contrasts["video"] = contrasts["clicDvideo"] + contrasts["clicGvideo"] + \
+    contrasts["calculvideo"] + contrasts["phrasevideo"]
+contrasts["computation"] = contrasts["calculaudio"] + contrasts["calculvideo"]
+contrasts["sentences"] = contrasts["phraseaudio"] + contrasts["phrasevideo"]
+
+# Short list or more relevant contrasts
+# contrasts = {
+#     "left-right": (contrasts["clicGaudio"] + contrasts["clicGvideo"]
+#                    - contrasts["clicDaudio"] - contrasts["clicDvideo"]),
+#     "H-V": contrasts["damier_H"] - contrasts["damier_V"],
+#     "audio-video": contrasts["audio"] - contrasts["video"],
+#     "video-audio": -contrasts["audio"] + contrasts["video"],
+#     "computation-sentences": (contrasts["computation"] -
+#                               contrasts["sentences"]),
+#     "reading-visual": contrasts["phrasevideo"] - contrasts["damier_H"]
+#     }
+
+contrasts = {
+    "reading-visual": contrasts["phrasevideo"] - contrasts["damier_H"]
+    }
+
+### Permutation tests ###########################################
+
+# Design matrix obj arrangement for paradigm permutations
+# What about creating design_matrix class to wrap make design matrix and
+# respective parameters. Then could just pass that object
+
+design_matrix_obj = [frame_times, paradigm,
+                     {'hrf_model': 'canonical with derivative',
+                      'drift_model': 'cosine',
+                      'period_cut': 128}]
+
+# Run the permutations
+
+res = first_level_permutation_test(contrasts, fmri_glm, fmri_img,
+                                   [design_matrix_obj], n_perm=600,
+                                   two_sided_test=True,
+                                   cluster_threshold=None,
+                                   random_state=None, verbose=10,
+                                   n_jobs=n_proc)
+
+### Plots #######################################################
+
+# Create snapshots of the contrasts thresholded with parametric and non
+# parametric methods.
+
+# write directory
+write_dir = 'results'
+if not path.exists(write_dir):
+    mkdir(write_dir)
+
+for index, (contrast_id, (unc_pval, cor_pval)) in enumerate(res.iteritems()):
+    z_map, = fmri_glm.transform(contrasts[contrast_id],
+                                contrast_name=contrast_id,
+                                output_z=True)
+    nib.save(z_map, path.join(write_dir, '%s_z_map.nii' % contrast_id))
+    z_map_data = fmri_glm.masker_.transform(z_map)[0]
+
+    display = plotting.plot_stat_map(z_map, display_mode='z',
+                                     threshold=z_score(0.1),
+                                     title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_z_map.png' % contrast_id))
+
+    nib.save(unc_pval, path.join(write_dir, '%s_unc_pval.nii' % contrast_id))
+    nib.save(cor_pval, path.join(write_dir, '%s_cor_pval.nii' % contrast_id))
+
+    threshold = (fmri_glm.masker_.transform(unc_pval)[0] < 0.1)
+    thresholded_z_map = fmri_glm.masker_.inverse_transform(z_map_data * threshold)
+
+    display = plotting.plot_stat_map(thresholded_z_map, display_mode='z',
+                                     threshold=0.5, title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_unc_pval.png' % contrast_id))
+
+    threshold = (fmri_glm.masker_.transform(cor_pval)[0] < 0.5)
+    thresholded_z_map = fmri_glm.masker_.inverse_transform(z_map_data * threshold)
+    display = plotting.plot_stat_map(thresholded_z_map, display_mode='z',
+                                     threshold=0.5, title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_cor_pval.png' % contrast_id))
--- a/nistats/permutation_tests.py
+++ b/nistats/permutation_tests.py
@@ -1,0 +1,296 @@
+"""
+Permutation tests for first level and second level analysis.
+"""
+# Author: Martin Perez-Guevara, <mperezguevara@gmail.com>, jan. 2016
+import warnings
+import numpy as np
+from sklearn.utils import check_random_state
+import sklearn.externals.joblib as joblib
+from sklearn.base import clone
+from nistats.design_matrix import make_design_matrix
+import time
+import sys
+
+
+def _meet_contrast_conditions(con_val):
+    """Check if permutation test is allowed for given contrast values."""
+    # No F tests
+    if type(con_val[0]) is list:
+        return False
+    # Not with less or more than 2 conditions
+    if len(np.where(con_val != 0.0)[0]) != 2:
+        return False
+    return True
+
+
+def _permutations_glm(original_stat, glm_ref, imgs, design_matrix_objs,
+                      con_val, con_id, con_labels, n_perm_chunk=10000,
+                      two_sided_test=True,
+                      random_state=None,
+                      n_perm=10000,
+                      thread_id=1,
+                      verbose=0):
+    """Massively univariate group analysis with permuted OLS on a data chunk.
+    To be used in a parallel computing context.
+    Parameters
+    ----------
+    design_matrix_objs: list of [time_frames, paradigm, kwargs dict].
+        frame_times is an array of shape (n_frames,) representing the timing
+        of the scans in seconds. Paradigm is a DataFrame instance with the
+        description of the experimental paradigm. kwargs are any other
+        arguments that could be passed to the function make_design_matrix
+        defined in nistats.design_matrix.
+
+    original_stat : array-like, shape=(n_voxels)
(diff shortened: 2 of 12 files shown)


MartinPerez:
Cluster statistics would be easily included in line 112. I think it would be good to wait for #35 thresholding api to be confirmed to develop it.

MartinPerez:
For second level permutation I think I could fetch the haxby dataset (6 subjects) and perform the test on the well known object vs faces contrast?

bthirion:
For second-level analysis, I would take the localizer dataset and perform a one-sample test on a given contrast, then possibly a two-sample tests, e.g. where you compare two populations (e.g. male vs female) , or two contrasts (e.g. right motor versus left motor activity).

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+
+More specifically:
+
+1. A sequence of fMRI volumes are loaded
+2. A design matrix describing all the effects related to the data is computed
+3. a mask of the useful brain volume is computed
+4. A GLM is applied to the dataset (effect/covariance,
+   then contrast estimation)
+5. Permutation tests are performed to obtain uncorrected and corrected p values
+at each voxel for a contrast comparing only two conditions.
```
bthirion: 'comparing two conditions' (no 'only')

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+
+from nistats.glm import FirstLevelGLM
+from nistats.design_matrix import make_design_matrix
+from nistats import datasets
+from nistats.permutation_tests import first_level_permutation_test
+from nistats.utils import z_score
+import nibabel as nib
+
+
+n_proc = 6
```
bthirion: should be `n_jobs=1`
(examples should work with n_jobs=1, since people will often run them on a laptop).

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+#     "left-right": (contrasts["clicGaudio"] + contrasts["clicGvideo"]
+#                    - contrasts["clicDaudio"] - contrasts["clicDvideo"]),
+#     "H-V": contrasts["damier_H"] - contrasts["damier_V"],
+#     "audio-video": contrasts["audio"] - contrasts["video"],
+#     "video-audio": -contrasts["audio"] + contrasts["video"],
+#     "computation-sentences": (contrasts["computation"] -
+#                               contrasts["sentences"]),
+#     "reading-visual": contrasts["phrasevideo"] - contrasts["damier_H"]
+#     }
+
```
bthirion: You can safely remove lines 68-87.

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+contrasts = {
+    "reading-visual": contrasts["phrasevideo"] - contrasts["damier_H"]
+    }
+
+### Permutation tests ###########################################
+
+# Design matrix obj arrangement for paradigm permutations
+# What about creating design_matrix class to wrap make design matrix and
+# respective parameters. Then could just pass that object
+
```
bthirion: As far as I can see, you don't need an object here, because you are running a computation, and that there is nothing that you might further  re-use.

MartinPerez: What do you mean by not needing an object? that I could instead say contrasts['reading-visual'] = ....

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+from sklearn.base import clone
+from nistats.design_matrix import make_design_matrix
+import time
+import sys
+
+
+def _meet_contrast_conditions(con_val):
+    """Check if permutation test is allowed for given contrast values."""
+    # No F tests
+    if type(con_val[0]) is list:
```
MartinPerez: I think this check is wrong...

examples/plot_localizer_analysis_permutation.py
bthirion: But then, what you need is simply a dictionary ?
Maybe we should discuss this more directly.

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+    display = plotting.plot_stat_map(thresholded_z_map, display_mode='z',
+                                     threshold=0.5, title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_unc_pval.png' % contrast_id))
+
+    threshold = (fmri_glm.masker_.transform(cor_pval)[0] < 0.5)
+    thresholded_z_map = fmri_glm.masker_.inverse_transform(z_map_data * threshold)
+    display = plotting.plot_stat_map(thresholded_z_map, display_mode='z',
+                                     threshold=0.5, title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_cor_pval.png' % contrast_id))
```
bthirion: display.show() at the end

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+
+    nib.save(unc_pval, path.join(write_dir, '%s_unc_pval.nii' % contrast_id))
+    nib.save(cor_pval, path.join(write_dir, '%s_cor_pval.nii' % contrast_id))
+
+    threshold = (fmri_glm.masker_.transform(unc_pval)[0] < 0.1)
+    thresholded_z_map = fmri_glm.masker_.inverse_transform(z_map_data * threshold)
+
+    display = plotting.plot_stat_map(thresholded_z_map, display_mode='z',
+                                     threshold=0.5, title=contrast_id)
+    display.savefig(path.join(write_dir, '%s_unc_pval.png' % contrast_id))
```
bthirion: this is  a misnomer because the image displays z values, not p-values.

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+
+for index, (contrast_id, (unc_pval, cor_pval)) in enumerate(res.iteritems()):
+    z_map, = fmri_glm.transform(contrasts[contrast_id],
+                                contrast_name=contrast_id,
+                                output_z=True)
+    nib.save(z_map, path.join(write_dir, '%s_z_map.nii' % contrast_id))
+    z_map_data = fmri_glm.masker_.transform(z_map)[0]
+
+    display = plotting.plot_stat_map(z_map, display_mode='z',
+                                     threshold=z_score(0.1),
```
bthirion: you're a bit extreme in the choice of the threshold. For an uncorrected  p-value, rather take .001

examples/plot_localizer_analysis_permutation.py
```diff
@@ -0,0 +1,148 @@
...
+                                     title=contrast_id)
```
bthirion: the titles should be different for the 3 figures

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        return False
+    # Not with less or more than 2 conditions
+    if len(np.where(con_val != 0.0)[0]) != 2:
```
bthirion: if np.sum(con_val != 0) != 2:

bthirion: bzw, you're assuming a t contrast. This should be checked.

nistats/permutation_tests.py
bthirion: Yes, this should checked by other means. 
I think that the API would clearer if you were providing the identifiers (strings) of two columns of the design matrix on the difference of which you want to make inference.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        return False
+    return True
+
+
+def _permutations_glm(original_stat, glm_ref, imgs, design_matrix_objs,
+                      con_val, con_id, con_labels, n_perm_chunk=10000,
+                      two_sided_test=True,
+                      random_state=None,
+                      n_perm=10000,
+                      thread_id=1,
```
bthirion: what is this thread_id ?

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+    if len(np.where(con_val != 0.0)[0]) != 2:
+        return False
+    return True
+
+
+def _permutations_glm(original_stat, glm_ref, imgs, design_matrix_objs,
+                      con_val, con_id, con_labels, n_perm_chunk=10000,
+                      two_sided_test=True,
+                      random_state=None,
+                      n_perm=10000,
```
bthirion: n_perm and n_perm_chunks are redundant.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+                      thread_id=1,
+                      verbose=0):
+    """Massively univariate group analysis with permuted OLS on a data chunk.
+    To be used in a parallel computing context.
+    Parameters
+    ----------
+    design_matrix_objs: list of [time_frames, paradigm, kwargs dict].
+        frame_times is an array of shape (n_frames,) representing the timing
+        of the scans in seconds. Paradigm is a DataFrame instance with the
```
bthirion: paradigm

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        description of the experimental paradigm. kwargs are any other
+        arguments that could be passed to the function make_design_matrix
+        defined in nistats.design_matrix.
+
+    original_stat : array-like, shape=(n_voxels)
```
bthirion: Do you need to give this ? This should be computed inside the function

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+
+    random_state : int or None,
+        Seed for random number generator, to have the same permutations
+        in each computing units.
+
+    Returns
+    -------
+    unc_ranks_parts: array-like, shape=(n_voxels)
+        Accumulated count of permuted stats over original stats,
+        to compute uncorrected p-values. (limited to this permutation chunk).
```
bthirion: The notion of permutation chunk is unclear at that point.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        If False, only positive effects are considered as relevant. The null
+        hypothesis is that the effect is zero or negative.
+
+    random_state : int or None,
+        Seed for random number generator, to have the same permutations
+        in each computing units.
+
+    Returns
+    -------
+    unc_ranks_parts: array-like, shape=(n_voxels)
```
bthirion: 'uncorrected_ranks_part'
Could you use a more explicit name ?

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        Accumulated count of permuted stats over original stats,
+        to compute uncorrected p-values. (limited to this permutation chunk).
+
+    smax_parts : array-like, shape=(n_perm_chunk, )
```
bthirion: maybe `t_max_chunks`

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+            perm_labels = rng.permutation(sel_labels['name'].tolist())
+            for lidx, perm_label in zip(labels_idx, perm_labels):
+                perm_paradigm.loc[lidx, 'name'] = perm_label
+
+            design_matrices.append(make_design_matrix(frame_times,
+                                                      perm_paradigm,
+                                                      **kwargs))
+
+        # Get permuted stat
+        glm_obj = clone(glm_ref)
```
bthirion: You should probably not clone.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+                    "Job #%d, processed %d/%d permutations for contrast %s "
+                    "(%0.2f%%, %i seconds remaining)%s"
+                    % (thread_id, i, n_perm_chunk, con_id, percent, remaining,
+                       crlf))
+
+    return unc_rank_parts, smax_parts
+
+
+def first_level_permutation_test(contrasts, glm_ref, imgs,
+                                 design_matrix_objs, n_perm=50,
```
bthirion: `n_perm=50` is not  a reasonable default for a number of permutations. I would take 1000

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+        permuted_stat, = glm_obj.fit(imgs, design_matrices) \
+                                .transform(con_val, contrast_name=con_id,
+                                           output_z=False, output_stat=True)
+        permuted_stat = glm_obj.masker_.transform(permuted_stat)[0]
+
+        # For uncorrected p-values
+        if two_sided_test:
+            unc_rank_parts += (np.fabs(permuted_stat) < original_stat)
+        else:
+            unc_rank_parts += (permuted_stat < original_stat)
```
bthirion: I thinks that storing the permutation values instead of the result of the threshold would avoid requiring the original stat in this function and would make it easier to understand: it would simply be a generator of statistical values to compare with an unknwon reference. The test would be handled in another function.
Try to keep all the functions as specific and simple as possible.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+    conditions = np.where(con_val != 0.0)
+    t0 = time.time()
+    for i in range(n_perm_chunk):
+        # Shuffle labels of the conditions in the design matrix
+        design_matrices = []
+
+        for frame_times, paradigm, kwargs in design_matrix_objs:
+            perm_paradigm = paradigm.copy()
+            con_idx = np.where(con_val != 0.0)[0]
+            conditions = [con_labels[x] for x in con_idx]
```
bthirion: Wouldn't it make more sense to provide those as arguments, rather than con_val ?

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+            selection = np.array([False for x in range(len(paradigm))])
+            for condition in conditions:
+                selection = selection | (perm_paradigm['name'] == condition)
```
bthirion: `selection = [x in conditions for x in perm_paradigm['name']]` is clearer to me

nistats/permutation_tests.py
bthirion: Note that providing a list of identifiers rather than a contrast would actuallay make it possible to permute more than 2 items.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+            warnings.warn('Contrast %s do not meet conditions.'
+                          'Computation ignored.' % (con_id,))
+            continue
+
+        # Get original stat
+        design_matrices = []
+        for time_frames, paradigm, kwargs in design_matrix_objs:
+            design_matrices.append(make_design_matrix(time_frames, paradigm,
+                                                      **kwargs))
+        glm_obj = clone(glm_ref)
```
bthirion: Cloning should be avoided

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+                          'set to %s. This is incompatible so only %d jobs will '
+                          'be running. You may want to perform more permutations '
+                          'in order to take the most of the available computing '
+                          'resources.' % (n_perm, n_jobs, n_perm))
+            n_perm_chunks = np.ones(n_perm, dtype=int)
+
+        # initialize the seed of the random generator
+        rng = check_random_state(random_state)
+
+        per = joblib.Parallel(n_jobs=n_jobs, verbose=verbose)(
```
bthirion: `permutation_results` rather than `per`
The names should always be as explicit as possible.

nistats/permutation_tests.py
```diff
@@ -0,0 +1,296 @@
...
+              joblib.delayed(_permutations_glm)(
+                  original_stat, glm_ref, imgs, design_matrix_objs,
+                  con_val, con_id, design_matrices[0].columns.tolist(),
+                  n_perm_chunk=n_perm_chunk,
+                  two_sided_test=two_sided_test,
+                  random_state=rng.random_integers(np.iinfo(np.int32).max - 1),
+                  n_perm=n_perm, thread_id=thread_id, verbose=verbose)
+              for thread_id, n_perm_chunk in enumerate(n_perm_chunks))
+
+        unc_ranks_parts, smax_parts = zip(*per)
```
bthirion: Same here regarding naming conventions.

bthirion:
Please add tests. This will help to detect potential bugs.

bthirion:
I'm going to look where the time is spent, as this business is rather slow.

bthirion:
A huge amount of time is spent masking the data. This should be done once and for all. Hence the permutations should directly performed on the masked data. Contrast computation also has a large overhead.

MartinPerez:
ok I will address all

MartinPerez:
this is now completely off the new objects and API and out of the roadmap. So I would close it if you agree @bthirion.

Issue #53, opened 2016-02-24
[enhancement] thresholding API should not require a mask


bthirion:

bthirion:
Solved. Closing.

PR #57, opened 2016-02-24
[MRG] Improved thresholding example

bthirion:
Some simplifications of the thresholding example + better doc. Addresses  #53


MartinPerez:
looks fine to me

PR #58, opened 2016-02-24
DOC: info update on dataset used in this ex.

alpinho:
add info about dataset used in example plot_spm_multimodal_faces.py


bthirion:
Great, thx ! Merging

Issue #70, opened 2016-05-30
Extreme memory consumption


MartinPerez:
I got the impression in my own datasets that memory consumption was extreme. I was consuming around 22 gb with only 5 runs corresponding to one hour of acquisition with 1.5 mm voxels. and started to get memory errors from other programs like ANTS, that is also memory hungry.

So I started profiling memory consumption of the fmri fit within the examples:

Here is memory consumption for the simple localizer example that loads a 30mb nifti file. the fmri fit explodes to 500mb after having only 100mb in memory. 

![nistats_memprofile_30-05-2016](https://cloud.githubusercontent.com/assets/8766915/15658262/9d4e1e82-26ba-11e6-9036-e5a3312cb9fb.png)

Here is the profile for the fiac analysis example

![nistats_memprofile2_30-05-2016](https://cloud.githubusercontent.com/assets/8766915/15658503/6eab723e-26bd-11e6-897a-75d79e547e87.png)

I think this has to be handled with high priority. Likely in my latest dataset with 16 runs over 2 sessions it will crash. Will test it soon.

bthirion:
I think that this one has been solved

Issue #44, opened 2015-12-03
Mean scaling conventions and options


MartinPerez:
I am not sure about all the implications of mean scaling the signal by voxels or globally, but since in nistats we change the conventions of SPM, maybe it would be a good idea to open this as an option instead of forcing the user to follow a different convention? Also for matters of replicability of SPM results.

In glm.py in the documentation we have:
 Y : array of shape(n_time_points, n_voxels)
then in line 54:
mean = Y.mean(0).

What do you think?

bthirion:
Correct me if I'm wrong, but the SPM standard consists in replacing the voxel-level mean signal. with the image mean, i.e. mean = Y.mean(),
I have to say that I don't understand this because the percent signal change is quite different across voxels.
However, if you believe that it's important, I suggest to dedicate a PR to that, after checking what SPM does.

MartinPerez:
I think its important for reproducibility of results to leave the option open for the user to select it even if we adopt as default something more reasonable, and clarify it in the documentation. I remember being the global mean the case in SPM but honestly I dont understand all the implications, I think the main argument is to be able to meaningfully compare across voxels and ROIs since they share the same baseline, but also then there is the clear problem of mixing white and grey matter. I just think giving the option cost little and do not enter in conflict then with the standard imposed by what is still a main tool of analysis in fMRI.

bthirion:
Makes sense. Please issue a PR at your convenience.

PR #72, opened 2016-06-06
First level model

MartinPerez:
1. First level model object basic implementation ready. 

2. I added trivially modified examples newapi_localizer and newapi_fiac that will keep being modified as the roadmap evolves. To compare with real simple examples how things are changing. For now changes are minimal. The localizer example shows better that now the design matrix do not need to be precomputed, while the fiac example shows that previous behavior, passing directly design matrices, still works.

3. memory issues were handled by releasing unnecessary variables and allowing the user to get rid of some regression results that are not useful unless the model will be inspected afterwards in details. The parameter minimize_memory in FirstLevelModel object takes care of this.

Improvements over issue #70 can be appreciated

localizer

fiac

4. parallelization on OLS fit by voxel groups and on AR model by assigning each AR model to a job

5. issue #44 addressed by adding scaling_axis parameter. renamed percent_signal_change to signal_scaling.

6. Factored out contrasts functionality to their own script, for future coming improvements like formulas handling.

Note: glm.py is supposed to be completely replaced by first_level_model.py and contrasts.py. Its kept such that previous examples keep working.

Please let me know what you think.


MartinPerez:
@bthirion tests are still missing. Besides reusing glm.py tests, do you think I should test any new features in more details. For example different voxel scalings added?

MartinPerez:
Note: Only first_level_model.py, regression.py and utils.py require to be reviewed. The rest is moving code around. And the examples are almost identical to the original versions for now.

examples/newapi_fiac_analysis.py
```diff
@@ -0,0 +1,88 @@
+"""
```
bthirion: We should remove examples with the old API, or this will become a mess.

examples/newapi_localizer_analysis.py
```diff
@@ -0,0 +1,93 @@
+"""
```
bthirion: Same comment here. The example should inly use the best API, so please remove plot_localizer_analysis.py and rename newapi_localizer_analysis.py to plot_localizer_analysis.py.

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+"""
+This module presents an interface to use the glm implemented in
+nistats.regression.
+
+It contains the GLM and contrast classes that are meant to be the main objects
+of fMRI data analyses.
+
+It is important to note that the GLM is meant as a one-session General Linear
+Model. But inference can be performed on multiple sessions by computing fixed
+effects on contrasts
```
bthirion: This point no longer holds.

nistats/contrasts.py
```diff
@@ -0,0 +1,241 @@
+import numpy as np
```
bthirion: Please remove all the corresponding code from glm.py. I am quite anxious about code duplication.

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+
+from .regression import OLSModel, ARModel, concatenate_regression_results
+from .design_matrix import make_design_matrix
+from .contrasts import compute_contrast
+from .utils import _basestring
+from .utils import GroupIterator
+
+
+def mean_scaling(Y, axis=0):
+    """Scaling of the data to have percent of baseline change columnwise
```
bthirion: not columnwise, along the specified axis.

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+
+def _minimize_memory_regression_results(results):
+    del results.Y
+    del results.model
+    del results.wY
+    del results.wresid
+    return results
+
+
+def _OLSModel_fit(X, Y):
```
bthirion: _ols_model_fit

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+        warn('Mean values of 0 observed.'
+             'The data have probably been centered.'
+             'Scaling might not work as expected')
+    mean = np.maximum(mean, 1)
+    Y = 100 * (Y / mean - 1)
+    return Y, mean
+
+
+def _minimize_memory_regression_results(results):
+    del results.Y
```
bthirion: Please provide a one-line docstring

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+    del results.model
+    del results.wY
+    del results.wresid
+    return results
```
bthirion: I understand what youd o here, but I find it very dirty. Why not creating another simplified object ?

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+
+
+def _OLSModel_fit(X, Y):
+    return OLSModel(X).fit(Y)
+
+
+def _ARModel_fit(X, val, Y):
```
bthirion: _ar_model_fit

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+                format(acceptable_noise_models, noise_model))
+
+    if Y.shape[0] != X.shape[0]:
+        raise ValueError(
+            'The number of rows of Y should match the number of rows of X.'
+            ' You provided X with shape {0} and Y with shape {1}'.\
+                format(X.shape, Y.shape))
+
+    # Create the model
+    # model = OLSModel(X)
```
bthirion: remove this line

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+    # Parallelize model fit by voxel groups
+    n_voxels = Y.shape[1]
+    group_iter = GroupIterator(n_voxels, n_jobs)
```
bthirion: gloups, I don't know that guy, and I don't like his name.

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+
+    else:
+        labels = np.zeros(Y.shape[1])
+        results = {0.0: ols_result}
+
+    return labels, results
+
+
+def _check_list_length_match(list_1, list_2, var_name_1, var_name_2):
+    if len(list_1) != len(list_2):
```
bthirion: This function should have a (1-line) docstring and it should not be here, but in utils.py

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+        raise ValueError(
+            'len(%s) %d does not match len(%s) %d'
+            % (var_name_1, len(list_1), var_name_2, len(list_2)))
+
+
+def _check_and_load_tables(tables_, var_name):
```
bthirion: Same here

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+        elif isinstance(table, pd.DataFrame):
+            tables.append(table)
+        else:
+            raise TypeError('%s can only be a pandas DataFrames or a'
+                            'string. A %s was provided at idx %d' %
+                            (var_name, type(table), table_idx))
+    return tables
+
+
+def _check_run_tables(run_imgs, tables_, tables_name):
```
bthirion: idem

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+                                            self.fir_delays, confounds_matrix,
+                                            confounds_names, self.min_onset)
+            else:
+                design = design_matrices[run_idx]
+            self.design_matrices_.append(design)
+
+            # Compute GLM
+            Y = self.masker_.transform(run_img)
+
+            # Y = self.masker_.transform(run_img)
```
bthirion: Please remove these ghosts

nistats/first_level_model.py
```diff
@@ -0,0 +1,606 @@
...
+        -------
+        design_matrices : list of pandas DataFrames,
+            Holds the design matrices computed for the corresponding run_imgs
+        """
+        if self.design_matrices_ is None:
+            raise ValueError('The model has not been fit yet')
+        else:
+            return self.design_matrices_
+
+    def _all_runs_contrast(self, con_vals, stat_type):
```
bthirion: Please add a one-line docstring
Could this be a function instead of a method ?

nistats/first_level_model.py
bthirion: Is all this worth the added complexity for simple OLS fit ?

nistats/regression.py
```diff
@@ -330,3 +329,17 @@ def SSE(self):
...
+    theta = np.concatenate([result.theta for result in results], axis=1)
+    Y = np.concatenate([result.Y for result in results], axis=1)
+    model = results[0].model
+    wY = np.concatenate([result.wY for result in results], axis=1)
+    wresid = np.concatenate([result.wresid for result in results], axis=1)
+    dispersion = np.concatenate([result.dispersion for result in results],
+                                axis=0)
+    cov = results[0].cov
+    return RegressionResults(theta, Y, model, wY, wresid,
+                             dispersion=dispersion, cov=cov)
```
bthirion: I am a bit worried about np.concatenate, that can lead to unexpected behavior.
If all this is only used for OLS regression, I do not think that it is worth the effort.

bthirion:
There are some great API improvements ni this PR, thanks.
Could reduce code duplication whenever possible and test all the functions ? We cannot move forward without tests.

MartinPerez:
@bthirion thanks for the prompt feedback. I wanted your first feedback before eliminating all the other code. I will of course adapt the tests. I would agree with having a more simple regression results object and check afterwards if any extra complexity would be needed for other cases. I just wanted to show how the parallelization of OLS would look like, I will take it out and later test on big datasets the code to propose the change if its worth it.

MartinPerez:
@bthirion all comments addressed, all tests adapted and all examples adapted. This is ready for a second round of review

MartinPerez:
I added many more parameter checks and now compute contrast can take a single contrast definition for multiple runs. I think its safe to assume in that case we want the same contrast for all runs. Still is possible to provide one contrast per run to avoid including some of them.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+    Parameters
+    ----------
+
+    t_r: float
+        This parameter indicates repetition times of the experimental runs.
+        It is necessary to correctly consider times in the design matrix.
+        This parameter is also passed to nilearn.signal.clean.
+        Please see the related documentation for details.
+
+    slice_time_ref: float, optional (default 0.)
```
bthirion: The problem with your approach is that you no longer support irregular sampling, that was mad possible by directly providing frame_times. Leaving the possibility to use frame_times would yield now 3 arguments instead of 1. I understand that it makes the general case simpler (e.g. using BIDS convention that provide a TR and default slice_time, the frame_times would be inferred automatically).

nistats/contrasts.py
```diff
@@ -0,0 +1,261 @@
...
+            effect_[:, label_mask] = resl.effect
+            var_[:, :, label_mask] = resl.covariance
+
+    dof_ = regression_result[label_].df_resid
+    return Contrast(effect=effect_, variance=var_, dof=dof_,
+                    contrast_type=contrast_type)
+
+
+def _summary_contrast(labels, results, con_vals, stat_type):
+    """Computes the summary contrast assuming fixed effects.
```
bthirion: then it should be called _fixed_effect_contrast

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+    if (mean == 0).any():
+        warn('Mean values of 0 observed.'
+             'The data have probably been centered.'
+             'Scaling might not work as expected')
+    mean = np.maximum(mean, 1)
+    Y = 100 * (Y / mean - 1)
+    return Y, mean
+
+
+def _ar_model_fit(X, val, Y):
```
bthirion: I think we should avoid such one-liners. It makes debugging and reviewing harder and does not really help.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+        This parameter specifices the order of the drift model (in case it is
+        polynomial) for the design matrices.
+
+    fir_delays : array of shape(n_onsets) or list, optional
+        In case of FIR design, yields the array of delays used in the FIR
+        model.
+
+    min_onset : float, optional
+        This parameter specifies the minimal onset relative to the design
+        (in seconds). Events that start before (slice_time_ref * t_r +
```
bthirion: slice_time_ref + min_onset

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+
+    minimize_memory : boolean, optional
+        Gets rid of some variables on the model fit results that are not
+        necessary for contrast computation and would only be useful for
+        further inspection of model details. This has an important impact
+        on memory consumption. True by default.
+
+    subject_id : string, optional
+        id of the subject of this first_level_model instance. Will be used
+        to create subject specific memory cache if caching allowed.
```
bthirion: I don't think we need this feature. This signatures are becoming horrible, so I'd like to avoid that parameter.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+
+    target_shape: 3-tuple of integers, optional
+        This parameter is passed to nilearn.image.resample_img. Please see the
+        related documentation for details.
+
+    low_pass: False or float, optional
+        This parameter is passed to nilearn.signal.clean.
+        Please see the related documentation for details.
+
+    high_pass: False or float, optional
```
bthirion: parameters clash: we now also have a period_cut parameter that does the same thing.
I think that the masker should be called with low_pass/high_pass parameters set to False automatically, so that all the filtering is explicit in the design matrix. the low_pass and high_pass parameters would then disappear from this class constructor.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+    standardize : boolean, optional
+        If standardize is True, the time-series are centered and normed:
+        their variance is put to 1 in the time dimension.
+
+    signal_scaling: bool, optional,
+        If True, fMRI signals are scaled to the mean value of scaling_axis
+        Incompatible with standardize (standardize=False is enforced when\
+        signal_scaling is True).
+
+    scaling_axis: int or tuple, optional,
```
bthirion: signal_scaling and scaling_axis should become one single parameter, that would be one of {False, 0, 1, (0,1)}

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+
+        Parameters
+        ----------
+        ignore: None or list of strings, optional
+            Names of the parameters that are not returned.
+
+        Returns
+        -------
+        params: dict
+            The dict of parameters
```
bthirion: This docstring is not very informative (was already here before, maybe...).

bthirion: Oops, actually this function is never used. Please remove it.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+        """
+        return get_params(FirstLevelModel, self, ignore=ignore)
+
+    def fit(self, run_imgs, paradigms=None, confounds=None,
+            design_matrices=None):
+        """ Fit the GLM
+
+        For each run:
+        1. create design matrix
```
bthirion: 1. create design matrix X

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+                    "Computing run %d out of %d runs (%s)\r"
+                    % (run_idx, n_runs, remaining))
+
+            # Build the experimental design for the glm
+            run_img = check_niimg(run_img, ensure_ndim=4)
+            if design_matrices is None:
+                n_scans = run_img.get_data().shape[3]
+                if confounds is not None:
+                    confounds_matrix = confounds[run_idx].values
+                    if confounds_matrix.shape[0] != n_scans:
```
bthirion: this has laready been checked in l.379. Please remove.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+                        raise ValueError('Rows in confounds does not match'
+                                         'n_scans in run_img at index %d'
+                                         % (run_idx,))
+                    confounds_names = confounds[run_idx].columns
+                else:
+                    confounds_matrix = None
+                    confounds_names = None
+                start_time = self.slice_time_ref * self.t_r
+                end_time = (n_scans - self.slice_time_ref) * self.t_r
+                frame_times = np.linspace(start_time, end_time, n_scans)
```
bthirion: Here does the assumption of regular acquisition comes in. Since the only person I know who does not use such this scheme is your supervisor, I let you decide whether this is important or not ;-)

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+
+        Returns
+        -------
+        design_matrices : list of pandas DataFrames,
+            Holds the design matrices computed for the corresponding run_imgs
+        """
+        if self.design_matrices_ is None:
+            raise ValueError('The model has not been fit yet')
+        else:
+            return self.design_matrices_
```
bthirion: I think we should remove such getters that do exactly nothing.
writing model.design_matrices_ is not more complicated than model.get_design_matrices()

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+                                     ' array' % cidx)
+        else:
+            raise ValueError('contrast_def must be an array')
+        n_runs = len(self.labels_)
+        if len(con_vals) != n_runs:
+            warn('One contrast given, assuming it for all %d runs' % n_runs)
+            con_vals = con_vals * n_runs
+        if isinstance(output_type, _basestring):
+            if output_type not in ['z_score', 'stat', 'p_value', 'eff', 'var']:
+                raise ValueError('output_type must be "z_score", "stat",'
```
bthirion: must be one of

nistats/regression.py
```diff
@@ -330,3 +329,65 @@ def SSE(self):
     def MSE(self):
         """ Mean square (error) """
         return self.SSE / self.df_resid
+
+
+class SimpleRegressionResults(LikelihoodModelResults):
```
bthirion: We'll probably have to refactor this later, but for the moment I'm fine with it as is.

nistats/utils.py
```diff
@@ -7,10 +7,44 @@
 import numpy as np
 from scipy.stats import norm
 from warnings import warn
+import pandas as pd
 
 py3 = sys.version_info[0] >= 3
 
 
+def _check_list_length_match(list_1, list_2, var_name_1, var_name_2):
```
bthirion: Sorry, but all thee guys need to come with a test.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+        self.fir_delays = fir_delays
+        self.min_onset = min_onset
+        # glm parameters
+        self.mask = mask
+        self.target_affine = target_affine
+        self.target_shape = target_shape
+        self.low_pass = low_pass
+        self.high_pass = high_pass
+        self.smoothing_fwhm = smoothing_fwhm
+        if memory is not None:
```
bthirion: A test for that feature is welcome too.

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+            self.masker_ = NiftiMasker(
+                mask_img=self.mask, smoothing_fwhm=self.smoothing_fwhm,
+                target_affine=self.target_affine,
+                standardize=self.standardize, low_pass=self.low_pass,
+                high_pass=self.high_pass, mask_strategy='epi',
+                t_r=self.t_r, memory=self.memory,
+                verbose=max(0, self.verbose - 1),
+                target_shape=self.target_shape,
+                memory_level=self.memory_level)
+        else:
```
bthirion: The following lines are not covered by any test

nistats/first_level_model.py
```diff
@@ -0,0 +1,561 @@
...
+                else:
+                    remaining = (100. - percent) / max(0.01, percent) * dt
+                    remaining = '%i seconds remaining' % remaining
+                sys.stderr.write(
+                    "Computing run %d out of %d runs (%s)\r"
+                    % (run_idx, n_runs, remaining))
+
+            # Build the experimental design for the glm
+            run_img = check_niimg(run_img, ensure_ndim=4)
+            if design_matrices is None:
```
bthirion: The following lines are not covered by tests.

bthirion:
Ok, this is looking better and better. We just need a few more tests and remove some functions that are not really necessary imho and make the API a bit heavy.

MartinPerez:
I added many more tests, particularly for errors raised and contrast and utility functions. Also addressed all comments as we discussed too. The only thing still lacking a bit is more tests after l. 389.

MartinPerez:
@bthirion I think I added enough tests for all first level model methods and raised errors. Please let me know what you think.

MartinPerez:
interesting that the ar model was not giving the exact same ar partitions for labels.

MartinPerez:
@bthirion i dont understand why in python 3.4 AR models is do not have consistent results. Should I open an issue to take care of this later? since its not directly relevant to the api reformulation.

bthirion:
Hm, the  failure must be quite subtle, because I don't see by eye where the two vectors are different. In any case, we need to fix this travis failure.

MartinPerez:
@bthirion I am confident the failure is really unrelated with this PR. I did not manage to solve it quickly, I opened an issue and commented the part of the test that I added that highlighted the problem. Please lets put that aside to merge this after passing review. The second level object and contrast formulas are already finished in the next PR too.

nistats/contrasts.py
```diff
@@ -0,0 +1,262 @@
...
+            print('Automatically converted multi-dimensional t to F contrast')
+            contrast_type = 'F'
+        self.contrast_type = contrast_type
+        self.stat_ = None
+        self.p_value_ = None
+        self.baseline = 0
+        self.tiny = tiny
+        self.dofmax = dofmax
+
+    def eff(self):
```
bthirion: Do we need this ?

nistats/contrasts.py
```diff
@@ -0,0 +1,262 @@
...
+        return self.effect[0, :]
+
+    def var(self):
```
bthirion: Do we need this ?

nistats/contrasts.py
```diff
@@ -0,0 +1,262 @@
...
+        self.dim = effect.shape[0]
+        if self.dim > 1 and contrast_type is 't':
+            print('Automatically converted multi-dimensional t to F contrast')
+            contrast_type = 'F'
+        self.contrast_type = contrast_type
+        self.stat_ = None
+        self.p_value_ = None
+        self.baseline = 0
+        self.tiny = tiny
+        self.dofmax = dofmax
```
bthirion: I'm realizing the existence of this `dofmax` parameter. Apparently, it was already here before, but I'm quite worried that this is mostly meaningless and should probably be removed in the future.

nistats/first_level_model.py
```diff
@@ -0,0 +1,515 @@
...
+
+
+class FirstLevelModel(BaseEstimator, TransformerMixin, CacheMixin):
+    """ Implementation of the General Linear Model for single session fMRI data
+
+    Parameters
+    ----------
+
+    t_r: float
+        This parameter indicates repetition times of the experimental runs.
```
bthirion: in seconds.

nistats/first_level_model.py
```diff
@@ -0,0 +1,515 @@
...
+        This parameter specifies the cut period of the low-pass filter in
+        seconds for the design matrices.
+
+    drift_order : int, optional
+        This parameter specifices the order of the drift model (in case it is
+        polynomial) for the design matrices.
+
+    fir_delays : array of shape(n_onsets) or list, optional
+        In case of FIR design, yields the array of delays used in the FIR
+        model.
```
bthirion: in seconds

nistats/first_level_model.py
```diff
@@ -0,0 +1,515 @@
...
+        the design matrices. It can be 'canonical', 'canonical with derivative'
+        or 'fir'.
+
+    drift_model : string, optional
+        This parameter specifies the desired drift model for the design
+        matrices. It can be 'polynomial', 'cosine' or 'blank'.
+
+    period_cut : float, optional
+        This parameter specifies the cut period of the low-pass filter in
+        seconds for the design matrices.
```
bthirion: in seconds

nistats/first_level_model.py
bthirion: I undertsand that the check is different, but please add inline comments to help readers.

nistats/tests/test_dmtx.py
```diff
@@ -159,59 +157,29 @@ def test_design_matrix0d():
...
-def test_design_matrix1():
-    # basic test based on basic_paradigm and glover hrf
-    tr = 1.0
-    frame_times = np.linspace(0, 127 * tr, 128)
-    paradigm = basic_paradigm()
-    hrf_model = 'glover'
-    X, names = design_matrix_light(frame_times, paradigm, hrf_model=hrf_model,
-                            drift_model='polynomial', drift_order=3)
-    assert_equal(len(names), 7)
-
```
bthirion: Why do you remove this test ?

nistats/tests/test_dmtx.py
```diff
@@ -159,59 +157,29 @@ def test_design_matrix0d():
...
-def test_design_matrix1c():
-    # idem test_design_matrix1, but different test
-    tr = 1.0
-    frame_times = np.linspace(0, 127 * tr, 128)
-    paradigm = basic_paradigm()
-    hrf_model = 'glover'
-    X, names = design_matrix_light(frame_times, paradigm, hrf_model=hrf_model,
-                        drift_model='polynomial', drift_order=3)
     assert_true((X[:, - 1] == 1).all())
```
bthirion: Same question here.

nistats/tests/test_first_level_model.py
```diff
@@ -0,0 +1,308 @@
...
+                assert_raises(
+                    ValueError, FirstLevelModel(mask=None, t_r=1.0).fit, fi, d)
+                assert_raises(
+                    ValueError, FirstLevelModel(mask=None, slice_time_ref=0.).fit, fi, d)
+            # confounds rows do not match n_scans
+            assert_raises(
+                ValueError, FirstLevelModel(mask=None).fit, fi, d, conf)
+
+
+def basic_paradigm():
```
bthirion: This test should not be here but in test_dmtx.py or test_experimental_paradigm.py

nistats/tests/test_first_level_model.py
```diff
@@ -0,0 +1,308 @@
...
+                                minimize_memory=False)
+        model = model.fit(func_img, paradigm)
+        labels1 = model.labels_[0]
+        results1 = model.results_[0]
+        labels2, results2 = run_glm(model.masker_.transform(func_img),
+                                    model.design_matrices_[0], 'ar1')
+        # ar not giving consistent results in python 3.4
+        # assert_almost_equal(labels1, labels2, decimal=1) ####FIX
+        # assert_equal(len(results1), len(results2)) ####FIX
+
```
bthirion: See issue #77

nistats/contrasts.py
bthirion: Opened issue #79.

bthirion:
Aside from minor points, this is almost OK.

nistats/contrasts.py
MartinPerez: I use to simplify the code in compute_contrast, since after getting the contrast object I can call any method like 'stat' or 'p_value', including then 'eff' and 'var'.

nistats/contrasts.py
MartinPerez: same

nistats/tests/test_first_level_model.py
MartinPerez: basic paradigm is an utility function used by the tests, not a test itself.

nistats/tests/test_dmtx.py
MartinPerez: all the matrix1* tests are now together in test_design_matrix1 in line 172. There were basically 4 tests duplicating all lines of code except the last one. I did not see the point of having 4 separate tests.

nistats/tests/test_dmtx.py
MartinPerez: same as above

MartinPerez:
@bthirion I updated the documentation with seconds and answered your comments.

nistats/contrasts.py
bthirion: then can them `effect()` and `variance()`.
If they are bound to be used only there, to avoid redundancy with the class parameters, I would have a unique `summary_statistics()` method that returns both.

nistats/tests/test_dmtx.py
bthirion: OK.

nistats/contrasts.py
MartinPerez: what I mean is that in l. 547 of first_level_model.py I call the output_type as an attribute of the contrast object to simplify the code. A summary_statistics() method would make me revert to the switch logic that existed before I made this change. I thought this way was cleaner than handling each possible output_type with an if statement. I also thought of using the @setattr_on_read tag on the z_score, p_value and stat methods instead of creating the dummy functions. What do you prefer?

bthirion: I see.
* for the moment, rename `eff` to `effect` at least or, even better, 'effect_size'. Similarly 'var' should be 'variance'.
* add an inline comment to explain the purpose of this function e.g. 'make access to summary statistics more straightforward when computing contrasts'
*  @setattr_on_read that makes sense. I have no strong feeling about it.

MartinPerez:
@bthirion updated the contrasts interface, compute contrast method and documentation and tests accordingly.

nistats/first_level_model.py
```diff
@@ -0,0 +1,516 @@
...
+                    if confounds_matrix.shape[0] != n_scans:
+                        raise ValueError('Rows in confounds does not match'
+                                         'n_scans in run_img at index %d'
+                                         % (run_idx,))
+                    confounds_names = confounds[run_idx].columns
+                else:
+                    confounds_matrix = None
+                    confounds_names = None
+                start_time = self.slice_time_ref * self.t_r
+                end_time = (n_scans - self.slice_time_ref) * self.t_r
```
bthirion: Actually 
`end_time = (n_scans - 1 + self.slice_time_ref) * self.t_r`

bthirion:
Could you take care of this last fix (one line), then I'll merge ?

MartinPerez:
@bthirion done

bthirion:
All right, merging.

Issue #71, opened 2016-06-06
Parametric modulation - differences between packages


rschmaelzle:
Hi, 
we are contemplating to switch more and more to nistats and are thus checking for correspondence between nistats-results and existing methods (matlab, spm ...). So far the results have been very promising and I'm finding e.g. a correlation of >.96 between a SPM-ROI result across a sample of 20something participants and the nistats results. 

However, I am running into trouble when looking at a parametric modulation. I am use a block-regressor for modeling some visual response, and then use a separate parametric modulator (in my testcase it is just an ascending z-transformed vector that ranges from 1:nstimuli - [see examples of the design mat](https://drive.google.com/folderview?id=0B-sCgoERRQPaT25jQlBWMEtEdU0&usp=sharing). Here the correlation between SPM and nistats goes down to only .65. - .7. Comparing the design matrices and regressors that get built shows that they are pretty much identical. It seems that the diverging results come in somewhere in how the model gets estimated (collinearity, orthogonalization, type of regression etc?). Before going down into rabbitholes here, I wanted to ask if anyone had a quick idea or (ideally) knew some command that would bring results into correspondence.
Thanks, Ralf.

bthirion:
Dear Ralf,

Thanks for sharing this issue. No clear idea on that. I have to admit that we have not tested very thoroughly the parametric modulations in nistats so far. So the culprit is likely on our side.
As a first check: what do you compare: t statistics or effects estimates ? I would expect the ensuing t statistics to be more consistent than effects estimate that depend on extrinsic parameter tuning.
Best.

Issue #65, opened 2016-02-25
API reformulation


MartinPerez:
A recent discussion pointed at several current weaknesses of the API:

+ The error prone need to reinput parameters. For example to instantiate design matrices at different pipeline steps.
+ Lack of separation between data and parameters. Like the relationship between TR and design. We might want to separate design_spec from design_matrix.
+ Misnaming of certain functions in the analysis process. Like the transform in the GLM object that can be misleading. Other functions might be added to the GLM object to get certain types of ouputs.
+ Too much historical baggage. We might want to experiment with different conventions. Like replacing model specification with contrast vectors by an R/Statsmodels type of specification.

The importance of getting the second level analysis object right was emphasized. There were some issues raised related to mask application at first or second level and how to handle complex designs (ex. factorial design). We have to balance carefully user surprisal with limiting the power of the framework for mere historical reasons.

Here a prototype proposal of API changes:

paradigm = load_paradigm('paradigm_sub1.csv')
confounds = pandas.read_csv('motion.csv')

model = GLM(hrf_model="spm", drift_model="", t_r=None,
            mask_img=False)
model.fit(paradigm, epi_imgs, confounds)
print(model.summary())
model.compute_contrast('seen - unseen')

second_level = SecondLevel(type="ffx", mask='auto')
second_level.fit([model1, model2, model3], design=None)
second_level.compute_contrast('(seen - unseen) * (schizo - control)')
second_level.compute_contrast(first='(seen - unseen)'
                              second='(schizo - control)')

bthirion:
Namy thx for the summary. Note that I'm working remotely today. I'll 
start by consider a bit mroe carefully patsy/statsmodel.
Best,

B

MartinPerez:
I will check this out too and bring some code proposal for march 7.

MartinPerez:
After discussion, @bthirion and me arrived to the API reformulation proposal posted in the wiki. https://github.com/nistats/nistats/wiki/API-proposal. Please give us your feedback!

Here a simple uncommented example as teaser. Please check the wiki for detailed comments on the proposal.


`design_matrix_glm_kwargs = {'hrf_model': 'canonical', 'drift_model': 'cosine',`
`                            'low_pass': None, 'high_pass': None}`

`model_formula = 'y = cond1 + cond2 + cond3 + ar_noise'`
`firstlevel = FirstLevelModel('task_id', formula=model_formula, mask_img=None, n_jobs=n_proc,`  
[...]

PR #78, opened 2016-06-13
Second level model

MartinPerez:
Here is the second level model with one simple example (second level analysis only) on the mixed gambles dataset (which is more a sanity check than an actual example due to the lack of a nice contrast to picture the results). I will create a new PR for a complete first and second level example on a bids dataset. The second level model object is quite flexible at the moment, it accepts list of first level models, the pandas dataframe discussed with @bthirion  and a simple list of images, then the other arguments must be matched accordingly. I hope the documentation is clear in this respect.

At the moment the second level only includes a simple OLS regression that only takes into account effect maps (RFX). Future PRs would be done for more complex models.

I included the patsy formulas functionality for the contrasts, it was a very small adaptation of the code. So we already have formulas for contrasts in both first and second level objects.

Is built on top of the first level model, so it should be compared with #72 instead of master until #72 is merged.

All tests included too.


salma1601:
Hi @MartinPerez 
I guess this PR is a work in progress. Is it planned to be completed soon ?
I need second level pipeline and I am hesitating between nistats and spm

MartinPerez:
Hi @salma1601 actually the PR is ready and functional, in testing and review phase. With small typos like the one you noticed in the example due to the radical API changes. I am currently testing it with my own datasets to also notice if some additional API changes would be useful.

Being responsible I think if you need to do the analysis right now I would recommend to still use spm, in a couple of weeks this will likely be merged but even then it would be good to compare spm and nistats output. For example an issue was raised recently #71 showing that parametric modulators seem to be giving different results in both frameworks even though the rest is fine.

Please feel welcome to also jump in the discussion of the API changes. It would also be great if more people give a hand comparing packages with their datasets and user cases to make nistats more robust.

MartinPerez:
API discussion in #65

salma1601:
Thanks for the quick reply.
>It would also be great if more people give a hand comparing packages with their datasets and user cases to make nistats more robust.

OK I am going to launch first level (and second level as soon as merged) on my data and warn if problems

MartinPerez:
in first level model compute contrast added the option to get the fixed effect contrast by default or the list of corresponding run contrasts. This will make it easy to collect per run maps for machine learning or other purposes.

MartinPerez:
it seems there is an issue with pandas 0.12.0 with python 2.6 in the second level design matrix creation. checking it out...

bthirion:
On 19/06/2016 01:04, Martin Perez-Guevara wrote:
>
> in first level model compute contrast added the option to get the 
> fixed effect contrast by default or the list of corresponding run 
> contrasts. This will make it easy to collect per run maps for machine 
> learning or other purposes.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub 
> <https://github.com/nistats/nistats/pull/78#issuecomment-226969791>, 
> or mute the thread 
> <https://github.com/notifications/unsubscribe/AAOT1r7BSKnYH-W-MmEjWHk04ORJ1B1nks5qNHmBgaJpZM4Iz6-V>.
>
Could this be done in a separate PR ? It is very hard to ensure the 
consistency of the API and the abstractions used by this code, hence we 
need to separate the different problems / fixes / improvements.

MartinPerez:
ok i will separate the fixed effect feature in a different PR

MartinPerez:
@bthirion in the process of fixing the error on pandas version 0.12 (turns out enlargement with .loc was not supported by until 0.13) I started to test the minimum requirements and it turns out nilearn 0.1 create many import issues in the current code. Moreover nilearn 0.2 is checked for conflicts in conda such that it is not compatible with python 2.6.

In install.sh nilearn is installed directly with pip, but then this is not compatible with the minimum requirements we are announcing for nistats.

A simple annoying example is that the check_niimg api changed radically from v 0.1

I admit so far I was developing with the latest libraries installed and overlooking minimum requirements. I checked we need to do patsy 0.2.0 (support python 2.6) or then jump to 0.4.1 (first version to not announce conflicts with python 2.7 in conda)

What should we do about this? I would vote for following a latests nilearn version than 0.1 minimum requirements and going for pandas 0.13 as the minimum.

bthirion:
Yes, given the very early state of Nistats, it seems reasonable to support only recent versions of the dependencies. We don't want to deal with the burden of bw compatibility for pandas, nilearn etc.

MartinPerez:
ok, shall we also drop official support for python 2.6? many libraries do not seem to support it anymore in the conda dependencies check.

bthirion:
I would accept the same dependencies as Nilearn, i.e. Python 2.6 --for 
the moment.

MartinPerez:
travis needs to be restarted, some of the builds failed for connection problems or other unrelated travis errors.

continuous_integration/show-python-packages-versions.py
```diff
@@ -1,6 +1,7 @@
 import sys
 
-DEPENDENCIES = ['numpy', 'scipy', 'sklearn', 'matplotlib', 'nibabel', 'pandas']
+DEPENDENCIES = ['numpy', 'scipy', 'nilearn', 'nibabel', 'pandas', 'patsy',
+                'sklearn', 'matplotlib']
```
bthirion: Could all this be in a separate PR ?

examples/plot_mixed_gambles.py
```diff
@@ -0,0 +1,80 @@
...
+    nib.save(zmap, zmap_path)
+    zmap_paths.append(zmap_path)
+behavioral_target = np.ravel(['gain_%d' % (i + 1) for i in data.gain]).tolist()
+subjects_id = np.ravel([['sub_%02d' % (i + 1)] * 48 for i in range(16)])
+subjects_id = subjects_id.tolist()
+mask_filename = data.mask_img
+
+
+# Second level model #########################################
+# create first_level_input
```
bthirion: this comment is a bit misleading. "create input for second-level analysis" sounds better.

nistats/design_matrix.py
```diff
@@ -443,3 +443,31 @@ def plot_design_matrix(design_matrix, rescale=True, ax=None):
     ax.set_xticks(range(len(names)))
     ax.set_xticklabels(names, rotation=60, ha='right')
     return ax
+
+
+def _create_second_level_design(maps_table, confounds=None):
+    """Infers a second level design from a maps table.
+
+    The maps table is a pandas DataFrame with at least columns 'map_name' and
+    'subject_id', confounds is also a pandas DataFrame with at least two
```
bthirion: In some cases (only one map per subject), you don't want to have the subject specific regressors. I would not impose it.
Also, beware that 'second-level' could be some kind of multi-acquisition/multi-session setting.

examples/plot_mixed_gambles.py
```diff
@@ -0,0 +1,80 @@
...
+df_columns = [subjects_id, behavioral_target, zmap_paths]
+df_columns = zip(*df_columns)
+df_column_names = ['subject_id', 'map_name', 'effects_map_path']
+first_level_df = pd.DataFrame(df_columns, columns=df_column_names)
```
bthirion: I think that we should not call it `first_level_df`: this object actually represent the second-level inputs. Actually, looking at it a bit more, I find it confusing.
I think that  `map_name` is actually more something like `experimental_condition`.

nistats/design_matrix.py
```diff
@@ -443,3 +443,31 @@ def plot_design_matrix(design_matrix, rescale=True, ax=None):
...
+    columns 'subject_id' and one confounder.
+    """
```
bthirion: Also the `confounder` deserves clarification: conceptually, what is the conceptual difference between the confounder and the `maps_table` ?

nistats/first_level_model.py
```diff
@@ -406,7 +413,7 @@ def fit(self, run_imgs, paradigms=None, confounds=None,
                 start_time = self.slice_time_ref * self.t_r
                 end_time = (n_scans - 1 + self.slice_time_ref) * self.t_r
                 frame_times = np.linspace(start_time, end_time, n_scans)
-                design = make_design_matrix(frame_times, paradigms[run_idx],
+                design = make_design_matrix(frame_times, events[run_idx],
                                             self.hrf_model, self.drift_model,
```
bthirion: I think that I agree with the change from 'paradigms' to 'events', but please try and keep this type of changes for other PRs, since this has exactly no relationship with second-level models.

nistats/first_level_model.py
```diff
@@ -441,22 +448,27 @@ def fit(self, run_imgs, paradigms=None, confounds=None,
...
-
-        contrast_name : str, optional
-            name of the contrast
+            desired for all runs. The string can be a formula compatible with
+            the linear constraint of the Patsy library. Basically one can use
+            the name of the conditions as they appear in the design matrix of
+            the fitted model combined with operators /*+- and numbers.
+            Nonetheless all contrasts assume the baseline of the tested effect
+            as 0, so specifying a different baseline in a formula will have no
+            effect. Please checks the patsy documentation for formula examples:
```
bthirion: I am not sure that the baseline question should be discussed here (looks like a corner case).

nistats/first_level_model.py
```diff
@@ -467,23 +479,33 @@ def compute_contrast(self, contrast_def, contrast_name=None,
 
         Returns
         -------
-        output_image : Nifti1Image
-            The desired output image
+        output : Nifti1Image
+            The desired output image from a fixed effects computation.
```
bthirion: Why do emphasize the notion of fixed effect ? This seems a bit unnatural here.

nistats/first_level_model.py
```diff
@@ -467,23 +479,33 @@ def compute_contrast(self, contrast_def, contrast_name=None,
...
         elif isinstance(contrast_def, (list, tuple)):
             con_vals = contrast_def
-            for cidx, con in enumerate(contrast_def):
+        else:
+            raise ValueError('contrast_def must be an array or str or list of'
+                             ' (array or str)')
+        # We reinterpret formulas as vectors only if design has column names.
+        des_names = np.unique(self.design_matrices_[0].columns.tolist())
+        if len(des_names) == len(self.design_matrices_[0].columns):
+            di = DesignInfo(self.design_matrices_[0].columns.tolist())
```
bthirion: `design_info` instead of `di`

nistats/first_level_model.py
```diff
@@ -467,23 +479,33 @@ def compute_contrast(self, contrast_def, contrast_name=None,
...
-                    raise ValueError('contrast_def at index %i is not an'
-                                     ' array' % cidx)
+                    con_vals[cidx] = di.linear_constraint(con).coefs
         else:
-            raise ValueError('contrast_def must be an array or list of arrays')
+            for cidx, con in enumerate(con_vals):
+                if not isinstance(con, np.ndarray):
+                    raise ValueError('can not interpret formula when design'
+                                     ' do not contain columns with unique '
+                                     'names')
```
bthirion: 'Cannot interpret formula when design does not contain columns with unique identifiers'

bthirion: Hm, thinking a bit further: we certainly don't want design to contain columns with redundant identifiers, so this should be enforced in the design matrix construction, not here (I find this code is becoming too convoluted because we are taking care of such issues that should be handled elsewhere)

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+from patsy import DesignInfo
+
+from .first_level_model import FirstLevelModel
+from .regression import OLSModel, SimpleRegressionResults
+from .contrasts import compute_contrast
+from .utils import _basestring
+from .design_matrix import _create_second_level_design
+
+
+def run_glm(Y, X, n_jobs=1, verbose=0):
```
bthirion: We already have a run_glm in first-level_glm. As these two guys are doing the same thing, this one should be replaced by a call to the other one.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+        self.verbose = verbose
+        self.n_jobs = n_jobs
+        self.minimize_memory = minimize_memory
+        self.labels_ = None
+        self.results_ = None
+
+    def fit(self, first_level_input, first_level_conditions=None,
+            confounds=None, design_matrix=None,
+            first_level_conditions_name=None):
+        """ Fit the GLM
```
bthirion: Fit the second-level GLM

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+        1. create design matrix
+        2. do a masker job: fMRI_data -> Y
+        3. fit regression to (Y, X)
+
+        Parameters
+        ----------
+        first_level_input: list of `FirstLevelModel` objects or pandas
```
bthirion: as discussed in the example, I am afraid that this terminology will confuse users, as it can be interpreted "input to first level analysis". `second_level_input` makes more sense to me. Do you agree ?

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            If list of `FirstLevelModel` objects, then first_level_conditions
+            must be provided. If a pandas DataFrame, then must contain columns
+            subject_id, map_name, effects_map_path. If list of Niimg-like
+            objects then this will be taken literally as Y for the model fit
+            and design_matrix must be provided.
+
+        first_level_conditions: list of (str or array) or None
+            If first_level_input is a list of `FirstLevelModel` objects then
+            it is mandatory to provide a list of contrast definitions to pass
+            to the compute_contrast method of `FirstLevelModel` which can be
```
bthirion: avoid 'which can be', because it is not obvious what 'which' refers to in such nested sentences.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            a str or array. Check the compute_contrast documentation of
+            `FirstLevelModel` for more details on the contrast definitions.
+
+            If first_level_input is a pandas DataFrame then a list of strings
+            is expected with the names of maps that will be part of the model.
+            None indicates that all maps should be included.
```
bthirion: If None, all maps are included

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+            If first_level_input is a list of Niimg-like objects then this
+            argument will be ignored.
+
+        confounds: pandas DataFrame, optional
```
bthirion: I don't like this 'confounds' terminology, as this may well be an effect of interest (e.g. the age of the subject, see may point above). Use 'regressors' instead of 'confounds'.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            Must contain a subject_id column. All other columns will be
+            considered as confounders and included in the model. If
+            design_matrix is provided then this argument will be ignored.
+            The resulting second level design matrix will use the same column
+            names as in the given DataFrame for confounds. At least two columns
+            are expected, "subject_id" and at least one confounder.
```
bthirion: Nit-picking: avoid using future 'will' for these descriptions. This makes the formulation heavier for no benefit.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+        design_matrix: pandas DataFrame, optional
+            Design matrix that will be used to fit the GLM. The number of rows
+            in the design matrix must agree with the size of the list in
+            first_level_conditions. Careful that the order of maps given
```
bthirion: Be careful that

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            by first_level_conditions or inferred directly from
+            first_level_input when first_level_conditions is None correctly
+            match the order of the rows in the design matrix.
```
bthirion: Ensure that the order of maps given by first_level_conditions or inferred directly from first_level_input matches the order of the rows in the design matrix.

bthirion: Maybe we should think of mechanisms for such checks as this is not straightforward.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+        first_level_conditions_name: list of str or None,
+            This argument only applies when first_level_input is a list of
+            `FirstLevelModel` objects. It specifies the name of the columns
+            corresponding to the provided first_level_conditions.
```
bthirion: I find this a bit convoluted: this hints at a design problem. Esentially first_level_conditions_name and first_level_conditions point to the same object, this probably means that this one should be e.g. a dictionary

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            If None then the column names will be generated with the form
+            'con_00'.
+        """
+        # Check parameters
+        # check first level input
+        if isinstance(first_level_input, FirstLevelModel):
```
bthirion: This check is a bit weird. Why not checking that this is indeed a list of FirstleveModels, image or df ?

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            raise ValueError('A second level model requires a list with at'
+                             'least two first level models.')
+        elif isinstance(first_level_input, list):
+            if len(first_level_input) < 2:
+                raise ValueError('A second level model requires a list with at'
+                                 'least two first level models.')
+            # Check FirstLevelModel objects case
+            if isinstance(first_level_input[0], FirstLevelModel):
+                if first_level_conditions is None:
+                    raise ValueError('First level models input require'
```
bthirion: requires

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+                                         (midx, type(first_level_model)))
+                    if confounds is not None:
+                        if first_level_model.subject_id is None:
+                            raise ValueError('In case confounds are provided '
+                                             'first level objects need to '
+                                             'provide the attribute subject_id'
+                                             ' to match rows appropriately'
+                                             '. model at idx %d do not provide'
+                                             ' it. It can be set by doing '
+                                             'model.subject_id = "01"')
```
bthirion: first_level_model.subject_id = "01".

nitpicking: could you make longer lines by starting a new line after the opening parenthesis ?

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+        if isinstance(first_level_input, pd.DataFrame):
+            # Get only first level conditions if provided
+            if first_level_conditions is not None:
+                for condition in first_level_conditions:
+                    if condition not in first_level_input['map_name'].tolist():
+                        raise ValueError('condition %s not present in'
+                                         ' first_level_input' % condition)
+                in_cond = first_level_input.apply(lambda x: x['map_name'] in
+                                                  first_level_conditions,
+                                                  axis=1)
```
bthirion: Please avoid this: this makes the code uselessly hard to undertsand.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+                first_level_input = first_level_input[in_cond]
+            # Create design matrix if necessary
+            if design_matrix is None:
+                design = _create_second_level_design(first_level_input,
+                                                     confounds)
+            # get effect maps
+            effects_maps = first_level_input['effects_map_path'].tolist()
```
bthirion: Maybe this is not the right place for this comment, yet we should remember that for mixed-effects models, you typically want to have bit the first-level effects *and variance*.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            # Check models were fit
+            for midx, model in enumerate(first_level_input):
+                if model.labels_ is None:
+                    raise ValueError('Model at idx %d has not been fit' % midx)
+            # Get the first level model maps
+            design_df = pd.DataFrame(columns=['map_name', 'subject_id'])
+            effects_maps = []
+            for model in first_level_input:
+                for con_idx, con_def in enumerate(first_level_conditions):
+                    con_name = 'con_%02d' % con_idx
```
bthirion: should be first_level_condition_names rather than a rather arbitrary integer.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+                    if first_level_conditions_name:
```
bthirion: All right. I'd rather introduce a deafult for first_level conditions_name, but to me the right aprpoach is rather to use first_level_conditions.keys()

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+                        con_name = first_level_conditions_name[con_idx]
+                    design_df.loc[len(design_df)] = [con_name,
+                                                     model.subject_id]
+                    eff_map = model.compute_contrast(con_def,
+                                                     output_type='effect_size')
+                    effects_maps.append(eff_map)
+            # Get the design matrix
+            if design_matrix is None:
+                design = _create_second_level_design(design_df, confounds)
```
bthirion: I find it hard to undertsand the distinction between design_df and design_matrix.
* I think that the design of _create_second_levle _design is a key issue. This function should be part of the public API, and there has to be an example dedicated only to presenting what it does.
* Renaming will be important. `design` alone is not clear enough; `design_matrix` is certainly better. `design_df` is not great; may be `design_input`

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+        else:
+            design = design_matrix
+            effects_maps = first_level_input
+
+        # set design matrix, given or computed
+        if design_matrix is None:
+            self.design_matrix_ = design
+        else:
+            self.design_matrix_ = design = design_matrix
```
bthirion: :-1:

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+        Y = self.masker_.transform(effects_maps)
+        if self.memory is not None:
+            mem_glm = self.memory.cache(run_glm)
+        else:
+            mem_glm = run_glm
+        labels, results = mem_glm(Y, design.as_matrix(), n_jobs=self.n_jobs)
+        # We save memory if inspecting model details is not necessary
+        if self.minimize_memory:
+            for key in results:
+                results[key] = SimpleRegressionResults(results[key])
```
bthirion: Let's keep this for the moment, but in the future this will have to go.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            sys.stderr.write("\nComputation of second level model done in "
+                             "%i seconds\n" % (time.time() - t0))
+
+        return self
+
+    def compute_contrast(self, contrast_def, stat_type=None,
+                         output_type='z_score'):
+        """Generate different outputs corresponding to
+        the contrasts provided e.g. z_map, t_map, effects and variance.
+        In multi-session case, outputs the fixed effects map.
```
bthirion: this does not make sense here: no multi-session here (?)
Please update the whole docstring accordingly.

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+        if self.labels_ is None or self.results_ is None:
+            raise ValueError('The model has not been fit yet')
+
+        # check contrast definition
+        if isinstance(contrast_def, np.ndarray):
+            con_val = contrast_def
+            if np.all(con_val == 0):
+                raise ValueError('Contrast is null')
+        else:
+            des_names = np.unique(self.design_matrix_.columns.tolist())
```
bthirion: design_ames

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+            if len(des_names) == len(self.design_matrix_.columns):
+                di = DesignInfo(self.design_matrix_.columns.tolist())
```
bthirion: design_info

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+                con_val = di.linear_constraint(contrast_def).coefs
+            else:
+                raise ValueError('can not interpret formula when design'
```
bthirion: Should have been checked somewhere else

nistats/second_level_model.py
```diff
@@ -0,0 +1,451 @@
...
+
+        # check output type
+        if isinstance(output_type, _basestring):
+            if output_type not in ['z_score', 'stat', 'p_value', 'effect_size',
+                                   'effect_variance']:
+                raise ValueError('output_type must be one of "z_score", "stat"'
+                                 ' ,"p_value", "effect_size" or "effect_variance"')
+        else:
+            raise ValueError('output_type must be one of "z_score", "stat",'
+                             ' "p_value","effect_size" or "effect_variance"')
```
bthirion: missing space after comma

nistats/tests/test_first_level_model.py
```diff
@@ -296,6 +296,16 @@ def test_first_level_model_contrast_computation():
...
+        model.compute_contrast(c2, 't', 'p_value')
+        model.compute_contrast(c2, None, 'effect_size')
+        model.compute_contrast(c2, None, 'effect_variance')
+        # formula should work (passing varible name directly)
+        model.compute_contrast('c0')
+        model.compute_contrast('c1')
+        model.compute_contrast('c2')
         # smoke test for one null contrast in group
         model.compute_contrast([c2, cnull])
         # only passing null contrasts should give back a value error
```
bthirion: Could be in another PR ?

nistats/tests/test_second_level_model.py
```diff
@@ -0,0 +1,244 @@
...
+        mask, FUNCFILE, _ = write_fake_fmri_data(shapes)
+        FUNCFILE = FUNCFILE[0]
+        func_img = load(FUNCFILE)
+        # ols case
+        model = SecondLevelModel(mask=mask)
+        # asking for contrast before model fit gives error
+        assert_raises(ValueError, model.compute_contrast, [])
+        # fit model
+        Y = [func_img]*4
+        X = pd.DataFrame([[1]]*4)
```
bthirion: pep8...

bthirion:
That's all for now. We can discuss on design aspect. I think we will be mreo efficient if we can remove from this PR the aspects that are not related to second-level model: first-level model improvement and version things.
Thanks !

MartinPerez:
I fixed all comments, created an example and improved documentation of the second level design matrix and eliminated first_level_conditions_name by making first_level_conditions a list of tuples. I do not use a dictionary to be able to provide the conditions in a given order to match to design matrix rows, which is a proposed user case.

This depends on the dependencies and first_level_upgrades branches. When those are merged, I will rebase and update this PR.

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+        # Build the design matrix X and list of imgs Y for GLM fit
+        if isinstance(second_level_input, pd.DataFrame):
+            maps_table = second_level_input
+            # Get only first level conditions if provided
+            if first_level_conditions is not None:
+                for name, condition in first_level_conditions:
+                    if condition not in maps_table['map_name'].tolist():
+                        raise ValueError('condition %s not present in'
+                                         ' second_level_input' % condition)
+                condition_list = list(zip(*first_level_conditions))[:][1]
```
MartinPerez: this is the ugly way of handling zip in python 3 :S. Any advice?

MartinPerez:
@bthirion rebased done, all comments addressed and tests passed.

examples/plot_mixed_gambles.py
```diff
@@ -0,0 +1,81 @@
...
+# contrast estimation
+for index, (contrast_id, contrast_val) in enumerate(contrasts[:4]):
+    print('  Contrast % 2i out of %i: %s' %
+          (index + 1, len(contrasts), contrast_id))
+    z_map = second_level_model.compute_contrast(contrast_val,
+                                                output_type='z_score')
+
+    # Create snapshots of the contrasts
+    display = plotting.plot_stat_map(z_map, display_mode='z',
+                                     threshold=3.0, title=contrast_id)
```
bthirion: These maps are desperately empty. Isn't there a contrast that displays a significant effect ?

examples/plot_mixed_gambles.py
```diff
@@ -0,0 +1,81 @@
...
+subjects_id = subjects_id.tolist()
+mask_filename = data.mask_img
+
+
+# Second level model #########################################
+# create input for second-level analysis
+df_columns = [subjects_id, behavioral_target, zmap_paths]
+df_columns = zip(*df_columns)
+df_column_names = ['model_id', 'map_name', 'effects_map_path']
+first_level_df = pd.DataFrame(df_columns, columns=df_column_names)
```
bthirion: It seems clearer to me to call it `second_level_df`.

examples/plot_mixed_gambles.py
```diff
@@ -0,0 +1,81 @@
...
+subjects_id = np.ravel([['sub_%02d' % (i + 1)] * 48 for i in range(16)])
+subjects_id = subjects_id.tolist()
+mask_filename = data.mask_img
+
+
+# Second level model #########################################
+# create input for second-level analysis
+df_columns = [subjects_id, behavioral_target, zmap_paths]
+df_columns = zip(*df_columns)
+df_column_names = ['model_id', 'map_name', 'effects_map_path']
```
bthirion: 'map_name' sounds a bit strange: isn't it simply the experimental condition ?

examples/plot_second_level_design_matrix.py
```diff
@@ -0,0 +1,46 @@
...
+
+# Experimental paradigm has two conditions and 20 subjects
+column_names = ['map_name', 'model_id', 'map_path']
+n_subjects = 20
+maps_name = ['language'] * n_subjects + ['consonants'] * n_subjects
+subject_list = ['sub-%02d' % i for i in range(1, n_subjects + 1)]
+maps_model = subject_list * 2
+maps_path = [''] * n_subjects * 2
+maps_table = pd.DataFrame({'map_name': maps_name,
+                           'model_id': maps_model,
```
bthirion: I don't like `model_id`. Maybe rather `subject_id` ?

nistats/design_matrix.py
```diff
@@ -446,3 +446,47 @@ def plot_design_matrix(design_matrix, rescale=True, ax=None):
     ax.set_xticks(range(len(names)))
     ax.set_xticklabels(names, rotation=60, ha='right')
     return ax
+
+
+def create_second_level_design(maps_table, regressors=None):
+    """Infers a second level design from a maps table.
```
bthirion: Sets up a second level design...

nistats/design_matrix.py
```diff
@@ -446,3 +446,47 @@ def plot_design_matrix(design_matrix, rescale=True, ax=None):
...
+        information between them is matched based on the 'model_id' column
+        that both must have.
+
+    Returns
+    -------
+    design_matrix: pandas DataFrame
+        The second level design matrix
+    """
+    maps_name = maps_table['map_name'].tolist()
+    subjects_id = maps_table['model_id'].tolist()
```
bthirion: I think that we agree on renaming model_id to subject_id...

nistats/first_level_model.py
```diff
@@ -311,17 +312,17 @@ def fit(self, run_imgs, events=None, confounds=None,
...
+            Each column in a DataFrame corresponds to a regressor variable
             to be included in the regression model of the respective run_img.
             The number of rows must match the number of volumes in the
             respective run_img. Ignored in case designs is not None.
             If string, then a path to a csv file is expected.
 
         design_matrices: pandas DataFrame or list of pandas DataFrames,
             Design matrices that will be used to fit the GLM. If given it
-            takes precedence over events and confounds.
+            takes precedence over events and regressors.
```
bthirion: Why do you prefer `regressors` over `confounds` ? What I like with confounds is that it makes a clear distinction wrt the other regressors of interest.

MartinPerez:
Reverted model id to subject id and regressors to confounds. Changed example of mixed gambles to button press contrasts from localizer. I would add a BIDS example to show firstlevel-secondlevel more complex setup in the next BIDS utils PR.

MartinPerez:
the only point not addressed yet is the 'map_name', 'map_effect_size', 'map_effect_variance' names in the maps table. I think "map" is the most general term that is not misleading. Would you suggest a different word? Otherwise I handled what we discussed and latest comments.

examples/plot_second_level_button_press.py
```diff
@@ -0,0 +1,70 @@
...
+================================
+
+Full step-by-step example of fitting a GLM to perform a second level analysis
+in experimental data and visualizing the results.
+
+More specifically:
+
+1. A sequence of subject fMRI button press contrasts is downloaded.
+2. a mask of the useful brain volume is computed
+3. A GLM is applied to the dataset (as fixed effects, then contrast estimation)
```
bthirion: A one-sample t-test is applied to the brain maps.

examples/plot_second_level_button_press.py
```diff
@@ -0,0 +1,70 @@
...
+n_subjects = 16
+data = fetch_localizer_contrasts(["left vs right button press"], n_subjects,
+                                 get_tmaps=True)
+
+second_level_input = data['cmaps']
+design_matrix = pd.DataFrame([1] * len(second_level_input),
+                             columns=['left-right'])
+
+# Display subject t_maps
+subjects = [s[0] for s in data['ext_vars']]
```
bthirion: subject_data[0] for subject_data in data['ext_vars']

examples/plot_second_level_button_press.py
```diff
@@ -0,0 +1,70 @@
...
+                                            output_type='z_score')
+
+# Display group level z_map thresholded for p < 0.001
+p_val = 0.001
+z_th = norm.isf(p_val)
+display = plotting.plot_glass_brain(z_map, threshold=z_th, colorbar=True,
+                                    plot_abs=False, display_mode='z')
+display.savefig(os.path.join(write_dir, 'left-right_group.png'))
+
+plotting.show()
```
bthirion: This example is great, thanks.
Maybe another one that tests a differential effect (e.g. handedness) would help too.

examples/plot_second_level_design_matrix.py
```diff
@@ -0,0 +1,46 @@
...
+maps_name = ['language'] * n_subjects + ['consonants'] * n_subjects
+subject_list = ['sub-%02d' % i for i in range(1, n_subjects + 1)]
+maps_model = subject_list * 2
+maps_path = [''] * n_subjects * 2
+maps_table = pd.DataFrame({'map_name': maps_name,
+                           'subject_id': maps_model,
+                           'effects_map_path': maps_path})
+# Specify extra information about the subjects to create additional regressors
+extra_info_subjects = pd.DataFrame({'subject_id': subject_list,
+                                    'age': [i for i in range(15, 35)],
```
bthirion: `'age': range(15, 35),`
should work too.

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+    fMRI data
+
+    Parameters
+    ----------
+
+    mask: Niimg-like, NiftiMasker or MultiNiftiMasker object, optional,
+        Mask to be used on data. If an instance of masker is passed,
+        then its mask will be used. If no mask is given,
+        it will be computed automatically by a MultiNiftiMasker with default
+        parameters. Mask automatic computation assumes first level imgs have
```
bthirion: Automatic mask computation

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+    smoothing_fwhm: float, optional
+        If smoothing_fwhm is not None, it gives the size in millimeters of the
+        spatial smoothing to apply to the signal.
+
+    memory: string, optional
+        Path to the directory used to cache the masking process and the glm
+        fit. By default, no caching is done. Creates instance of joblib.Memory.
+
+    memory_level: integer, optional
+        Rough estimator of the amount of memory used by caching. Higher value
```
bthirion: This docstring is a bit hard to follow. When I read "Rough estimator of the amount of memory used by caching" I start thinking "My data are bout 300MB", which is not what is expected. The expected value is typically 1.

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+
+    verbose : integer, optional
+        Indicate the level of verbosity. By default, nothing is printed.
+        If 0 prints nothing. If 1 prints final computation time.
+        If 2 prints masker computation details.
+
+    n_jobs : integer, optional
+        The number of CPUs to use to do the computation. -1 means
+        'all CPUs', -2 'all CPUs but one', and so on.
+
```
bthirion: 'minimize_memory' should be documented.

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+        1. create design matrix
+        2. do a masker job: fMRI_data -> Y
+        3. fit regression to (Y, X)
+
+        Parameters
+        ----------
+        second_level_input: list of `FirstLevelModel` objects or pandas
+                            DataFrame or list of Niimg-like objects.
+            If list of `FirstLevelModel` objects, then first_level_conditions
+            must be provided. If a pandas DataFrame, then must contain columns
```
bthirion: ...then they have to contain...

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+            If second_level_input is a pandas DataFrame then a list with
+            column names as first item (employed in the design matrix) and their
+            corresponding map name as second item, where the map names correspond
+            to those in the second_level_input map_name column.
+            If first_level_conditions is set to None then all maps are included
+            and the map_name is used as the column names in the design matrix.
+
+            If second_level_input is a list of Niimg-like objects then this
+            argument is ignored.
+
```
bthirion: This docstring is also hard to read, because it describes a complex logic. This is where a series of examples for model specification will be useful

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+        if isinstance(second_level_input, list):
+            if len(second_level_input) < 2:
+                raise ValueError('A second level model requires a list with at'
+                                 'least two first level models or niimgs')
+            # Check FirstLevelModel objects case
+            if isinstance(second_level_input[0], FirstLevelModel):
+                if first_level_conditions is None:
+                    raise ValueError('First level models input requires'
+                                     'first_level_conditions to be provided')
+                for midx, first_level_model in enumerate(second_level_input):
```
bthirion: `midx` -> 'model_idx' would be more readable.

nistats/second_level_model.py
```diff
@@ -0,0 +1,381 @@
...
+
+        else:
+            effects_maps = second_level_input
+
+        # set design matrix, given or computed
+        self.design_matrix_ = design_matrix
+
+        # check design matrix X and effect maps Y agree on number of rows
+        if len(effects_maps) != design_matrix.shape[0]:
+            raise ValueError('design_matrix X number of rows do not agree with'
```
bthirion: does not match the number of maps considered

nistats/tests/test_second_level_model.py
```diff
@@ -0,0 +1,222 @@
...
+        c1, cnull = np.eye(ncol)[0, :], np.zeros(ncol)
+        # smoke test for different contrasts in fixed effects
+        model.compute_contrast(c1)
+        model.compute_contrast(c1, 'F')
+        model.compute_contrast(c1, 'F', 'z_score')
+        model.compute_contrast(c1, 'F', 'stat')
+        model.compute_contrast(c1, 'F', 'p_value')
+        model.compute_contrast(c1, 'F', 'effect_size')
+        model.compute_contrast(c1, 'F', 'effect_variance')
+        # formula should work (passing varible name directly)
```
bthirion: variable

bthirion:
All right, the PR looks good. With only tiny details remaining, it should be mergeable soon.

nistats/first_level_model.py
salma1601: Does this `confounds` plays the role of `add_regs` in the nipy design matrix ?

bthirion: Yes.

salma1601: Does it really need to be a confound ? For instance for ASL we can include additionnal regressors by convolving the PRF with the conditions. May be the terminology covariate can be used here ?

bthirion: The best answer is that we simply don't cover ASL for the moment. If you would like to contribute it, this should come as a specific PR. I think that it's necessary to have modality specific models if we want to obtain a user-friendly API.

salma1601: > If you would like to contribute it, this should come as a specific PR.

Ok, may be I can give that a try after some time.

>Note however that in spite of the 'confounds' term, these regressors are actually handled as well as the others in the glm and that you can make inference upon them if you wish or need.

Yes, that was the sens of my question. I found the word `confounds` a little misleading for it may suggest the opposite of what you are pointing to (actually it frightened me w.r.t. some old analyses, although I was almost sure all the regressors are treated the same way). To be honest I personally preferred the word `regressors`, or at least `covariates`. On the other hand I agree with modality specific models strategy.

examples/plot_second_level_button_press.py
MartinPerez: I would prefer to focus on sphinx, circle ci before developing more the tests. I will post today the barebones sphinx generated webpage.

nistats/second_level_model.py
MartinPerez: This is related to the masker and I am not sure myself of what the memory level is regulating. I agree the documentation is not helpful. This also appears in other parts of the code.

nistats/second_level_model.py
MartinPerez: I agree. Just opened an issue for this.

nistats/second_level_model.py
MartinPerez: I opened an issue for this.

nistats/second_level_model.py
MartinPerez: done

MartinPerez:
I addressed comments. I am posting today the sphinx circle ci branch to start handling previous and new examples well.

examples/plot_second_level_button_press.py
bthirion: OK, I'm going to open an issue for that.

bthirion:
I'll do another review today or tomorrow. We're close !

nistats/second_level_model.py
bthirion: No.

MartinPerez: tlooking again at this, would not be more readable:?

[condition[0] for condition in first_level_conditions]

no idea what passed my mind when I used zip here

nistats/design_matrix.py
```diff
@@ -446,3 +446,47 @@ def plot_design_matrix(design_matrix, rescale=True, ax=None):
...
+        if confounds is not None:
+            conrow = confounds['subject_id'] == row['subject_id']
+            for conf_name in confounds_name:
+                design_matrix.loc[ridx, conf_name] = confounds[conrow][conf_name].values
+
+    # check column names are unique
+    if len(np.unique(design_columns)) != len(design_columns):
+        raise ValueError('Design matrix columns do not have unique names')
+
+    return design_matrix
```
MartinPerez: so should I add a check like this here?:

linalg.cond(x) < 1/sys.float_info.epsilon

bthirion: yes. At this point I would probably spit a warning rather than an Error -- it is clear that using such a design matrix will lead to aberrant estimates.

MartinPerez:
I made the changes, need to kickstart travis.

MartinPerez:
my bad, travis is fine. All tests passed. circle ci is of course meant to fail in this branch.

bthirion:
All right, merging

PR #88, opened 2016-07-26
dependencies and travis update

MartinPerez:
This PR is simply to update the dependencies (Nilearn now only supports scikit learn 0.14.1 and we need some updates due to patsy 0.2.0 and pandas to 0.13). Moreover I added a travis server to test python 3.5.

### DIFF
--- a/continuous_integration/install.sh
+++ b/continuous_integration/install.sh
@@ -98,9 +98,11 @@ elif [[ "$DISTRIB" == "conda" ]]; then
         pip install nibabel=="$NIBABEL_VERSION"
     fi
     if [["$PYTHON_VERSION" == "2.6"]]; then
+        # Install the minimum supported version of nilearn and patsy
         pip install nilearn==0.2.0
         pip install patsy==0.2.0
     else
+        # Install the latest available version of nilearn and patsy
         pip install nilearn
         pip install patsy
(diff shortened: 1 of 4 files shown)


MartinPerez:
@bthirion this PR is ready for review.

continuous_integration/install.sh
```diff
@@ -94,6 +97,13 @@ elif [[ "$DISTRIB" == "conda" ]]; then
     if [ -n "$NIBABEL_VERSION" ]; then
         pip install nibabel=="$NIBABEL_VERSION"
     fi
+    if [["$PYTHON_VERSION" == "2.6"]]; then
```
bthirion: A comment is welcome here.

bthirion:
LGTM otherwise.

continuous_integration/install.sh
MartinPerez: something like "with python 2.6 we install the minimum version supported for patsy and nilearn otherwise the latest?"

MartinPerez: comments added under line 100. Please check in diff.

MartinPerez:
@bthirion comment added and tests passed.

bthirion:
Great, merging

PR #129, opened 2017-07-31
first level bids correction when sessions are present

MartinPerez:
A correction related to loading a BIDS dataset with sessions. There were some small issues and a print statement that was deleted.


bthirion:
Feels good, but a small test to enforce the behavior would be great  !

MartinPerez:
ok will add a test. Also several dependency chains seem broken, will look into.

bthirion:
Let me suggest to put all dependencies version changes in a separate PR, that would be merged first.

MartinPerez:
ok will separate them.

PR #130, opened 2017-08-22
version updates and related fixes due to pandas and linalg

MartinPerez:
There are some problems introduced by the sort function pandas being completely deprecated, now sort_values have to be used (introduced in pandas 0.17). There were some issues with linalg when passing matrices with dtype object (would not make automatic conversion to a numeric dtype)


MartinPerez:
@bthirion there are some issues due to mkl installation and libgfortran library installation after trying to upgrade to numpy 1.9, then neurodebian support numpy 1.8.2 while pandas sort function deprecation forces to upgrade to pandas 0.17.1 then forcing numpy to 1.9.

We would need to avoid using pandas sort function completely or upgrade to numpy 1.9 and related dependencies, what do you think is best?

MartinPerez:
for the moment i took the initiative to do without the pandas sort, since its only used in one line of code.

MartinPerez:
All travis problems fixed, @bthirion the version update is ready for review. The circle ci failure is related to a problem with the brainomics cea server, which is an issue also open in nilearn https://github.com/nilearn/nilearn/issues/678

bthirion:
Indee, I had come tio the same conclusion. I'll do a review in the afternoon, as we badly need these fixes.

MartinPerez:
Also I had to upgrade sphinx gallery to its latest release (0.1.12) to avoid another problem with circle ci. There was some folder naming convention broken with sphinx  1.6 (_extensions folder name changed to extentions). The new sphinx gallery version fixes the issue.

bthirion:
Aside from one small detail, LGTM.

MartinPerez:
what detail? we merge this?

bthirion:
Simply clarifying the pandas substitution (at least a comment, maybe an encapsulating function).

MartinPerez:
Done. I think a comment makes more sense for now, later we can create a function in utils if this comes up again, to make the compatibility constrain more explicit.

bthirion:
Thx. Merging

PR #134, opened 2017-08-30
Fix bids

MartinPerez:
fixes to get models from bids derivatives to work with openneuro datasets and new BIDS derivatives conventions. Will evolve as bugs appear during the coding sprint.

### DIFF
--- a/nistats/first_level_model.py
+++ b/nistats/first_level_model.py
@@ -566,7 +566,6 @@ class FirstLevelModel(BaseEstimator, TransformerMixin, CacheMixin):
         return output
 
 
-
 def first_level_models_from_bids(
         dataset_path, task_label, space_label, img_filters=[],
         t_r=None, slice_time_ref=0., hrf_model='glover', drift_model='cosine',
@@ -574,7 +573,7 @@ def first_level_models_from_bids(
         mask=None, target_affine=None, target_shape=None, smoothing_fwhm=None,
         memory=Memory(None), memory_level=1, standardize=False,
         signal_scaling=0, noise_model='ar1', verbose=0, n_jobs=1,
-        minimize_memory=True):
+        minimize_memory=True, derivatives_folder='derivatives'):
     """Create FirstLevelModel objects and fit arguments from a BIDS dataset.
 
     It t_r is not specified this function will attempt to load it from a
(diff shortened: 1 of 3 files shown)


nistats/first_level_model.py
```diff
@@ -743,7 +748,19 @@ def first_level_models_from_bids(
...
                                               img_dict['run']))
-                else:
+
+                elif '_ses-' in img_dict['file_basename']:
+                    if img_dict['ses'] in run_check_list:
+                        raise ValueError(
+                            'More than one nifti image found for the same ses '
+                            '%s, while no additional run specification present'
+                            '. Please verify that the preproc_variant and '
+                            'space_label labels were correctly specified.' %
```
bthirion: The message is fine, but you should relate this to the BIDS specs, otherwise the user may not understand what you're talking about (the users may not be fully aware of the BIDS stuff and simply use or test some features of Nistats).

nistats/tests/test_utils.py
```diff
@@ -133,6 +133,12 @@ def create_fake_bids_dataset(base_dir='', n_sub=10, n_ses=2,
                              tasks=['localizer', 'main'],
                              n_runs=[1, 3], with_derivatives=True,
                              with_confounds=True):
+    """Returns a fake bids dataset directory with dummy files
+
+    In the case derivatives are included, they come with two spaces and
+    variants. Spaces are 'MNI' and 'T1w'. Variants are 'some' and 'other'.
```
bthirion: you're probably referring to BIDS specs here ?

nistats/first_level_model.py
```diff
@@ -732,7 +736,8 @@ def first_level_models_from_bids(
         if len(imgs) > 1:
             for img in imgs:
                 img_dict = parse_bids_filename(img)
-                if '_ses-' in img_dict['file_basename']:
+                if ('_ses-' in img_dict['file_basename'] and
+                        '_run-' in img_dict['file_basename']):
```
bthirion: We should probably have tests covering these conditions.

nistats/tests/test_utils.py
MartinPerez: not really, they are dummy specs

MartinPerez:
added tests and small doc clarification on BIDS spec being referenced

MartinPerez:
rebased on download branch

nistats/datasets.py
```diff
@@ -93,46 +104,88 @@ def fetch_openfmri_dataset(dataset_name='ds000001', dataset_revision=None,
...
-    return data_dir, downloaded_files
+        with open(url_file, 'r') as json_file:
+            urls = json.load(json_file)
+
+    # We apply filters to the urls
+    for exclusion in exclusion_filters:
+        urls = [url for url in urls if not re.match(exclusion, url)]
+    for inclusion in inclusion_filters:
+        urls = [url for url in urls if re.match(inclusion, url)]
+
```
bthirion: I reiterate that with these conventions, exclusion and inclusion filter play logically the same role. Not a big deal though.

bthirion:
LGTM.
The big thing that I would like on top of it is an illustrative example.

bthirion:
LGTM.

PR #126, opened 2017-04-29
MRG: Better use of Nibabel

bthirion:
use .affine instead of .get_affine() etc.

PR #137, opened 2017-10-31
Nibabel use

bthirion:
This mostly uupgrades to curretn nibabel API. In place of #126

### DIFF
--- a/nistats/datasets.py
+++ b/nistats/datasets.py
@@ -342,7 +342,7 @@ def fetch_spm_auditory(data_dir=None, data_name='spm_auditory',
             vol = nibabel.load(x)
             if len(vol.shape) == 4:
                 vol = nibabel.Nifti1Image(vol.get_data()[:, :, :, 0],
-                                          vol.get_affine())
+                                          vol.affine)
                 nibabel.save(vol, x)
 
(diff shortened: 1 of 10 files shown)


nistats/datasets.py
```diff
@@ -339,22 +339,22 @@ def _glob_spm_auditory_data():
...
         # ... same thing for anat
-        vol = nibabel.load(_subject_data["anat"])
+        vol = nib.load(_subject_data["anat"])
         if len(vol.shape) == 4:
-            vol = nibabel.Nifti1Image(vol.get_data()[:, :, :, 0],
-                                      vol.get_affine())
-            nibabel.save(vol, _subject_data["anat"])
+            vol = nib.Nifti1Image(vol.get_data()[:, :, :, 0],
+                                  vol.affine)
+            vol.to_filename(_subject_data["anat"])
```
effigies: Here (and above) you may just want [`nib.squeeze_image`](https://github.com/nipy/nibabel/blob/2139ce0d24e65a83295bb6b3eaaf005eaeaebb5f/nibabel/funcs.py#L17-L87).

bthirion: All right, thx for the tip.

MartinPerez:
LGTM

Issue #151, opened 2017-11-28
clean webpage simple example upload


MartinPerez:
There is a file that should not be there in simple examples.

MartinPerez:
the old GLM fitting in second level fMRI example file was copied also by accident.

bthirion:
You mean, the plot_101 ?
OK, I'll remove it.

MartinPerez:
no, I mean in general examples the "GLM fitting in second level fMRI", that has to be removed, it is repeated in the second level section

bthirion:
Note sure I get what you mean. I think that keeping 
03_second_level_models/plot_second_level_button_press.py
is very important because it is something that can be reused easily, even without running the whole framework.

MartinPerez:
its not a problem with the repository, but the website. Simply re-uploading the website after merging #153 will fix the issue. If you open the website you will see that the second level example appears in two sections (in general examples and second level models).

kchawla-pi:
@GaelVaroquaux `make install` fails at the push stage, citing I don't have permission to do that.

bthirion:
Should be fine now. B

GaelVaroquaux:
I think that indeed the permissions were not set correctly on this repo, and the admin team did not have the rights. I fixed it.

kchawla-pi:
Fixed with #222

Issue #149, opened 2017-11-22
Authors of example should be moved to comments outside the docstring


GaelVaroquaux:
The reason is that in the docstring they then get turned into the summary that is printed in a gallery. The first sentence should be an information summary.

bthirion:
Solved by PR #153, closing.

PR #153, opened 2017-12-02
Clean examples

bthirion:
Addresses #149 and #151.


MartinPerez:
LGTM

bthirion:
Then, we should kill section 1 of the examples, that is now empty.
Or do you have any suggestion of an example that would fit well there ?

MartinPerez:
I think we should kill the section. Tutorials are themselves general examples anyway.

PR #155, opened 2018-01-02
Trying to fix circelci failure related to spm_auditory

bthirion:
Trying to fix circleci failure


bthirion:
CircleCI fixed, merging.

kchawla-pi:
- What was the problem? How did vol_to_file stop working? Was it replaced by new version of NIbabel that we use now? 
- Why was this not caught in the tests? Is it because of dataset caching? If so I will add another job to circleci, run occasionaly (once a day) to build without cache to catch such problems.

KamalakerDadi:
@kchawla-pi this thread might give you some details https://github.com/nipy/nibabel/issues/107

It is that nibabel loaded Nifti images cannot be save to files to the disk with extension .img, in particular using ```to_filename```. Whereas nibabel.save can whether it is .img or .nii.gz extension.

PR #161, opened 2018-01-28
Moved import boto3 to function level

### LINKED ISSUE #160: boto3 is acting as a dependency to install
boto3 is written instructions as to be installed if needed to download data from openneuro. But, without installing boto3 I could not able to import datasets module.

I think boto3 needs to be imported at function level than module level.

KamalakerDadi:
Fixes #160 

- Added helper function ```_check_import_boto3```.
- This helper function is used at two function relevant to openneuro
to ```import boto3``` and raises an error if not installed with a user friendly info.

### DIFF
--- a/nistats/datasets.py
+++ b/nistats/datasets.py
@@ -14,12 +14,26 @@ from sklearn.datasets.base import Bunch
 from nilearn.datasets.utils import (
     _get_dataset_dir, _fetch_files, _fetch_file, _uncompress_file)
 
-import boto3
 from botocore.handlers import disable_signing
 
 SPM_AUDITORY_DATA_FILES = ["fM00223/fM00223_%03i.img" % index
                            for index in range(4, 100)]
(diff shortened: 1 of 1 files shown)


nistats/datasets.py
```diff
@@ -14,14 +14,28 @@
...
+def _check_import_boto3(module_name):
+    """Helper function which checks boto3 is installed or not
+
+    If not installed raises an ImportError with user friendly
+    information.
+    """
+    try:
+        module = __import__(module_name)
+    except ImportError:
+        info = "Please install boto3 necessary to download openneuro datasets."
```
bthirion: Maybe: "Please install boto3 to download openneuro datasets."

bthirion: LGTM otherwise. Thx !

bthirion:
LGTM, thx.

PR #165, opened 2018-02-09
FIX: Remove global MPL state change

### LINKED ISSUE #164: Importing the reporting module changes appearance of unrelated plots
Libraries modifying matplotlib global state can make life difficult for downstream developers. Nistats does this at present:

https://github.com/nistats/nistats/blob/5745e7b6ffadc49b2c2b5c6948f4bb8a61bd7935/nistats/reporting.py#L16

Something I've looked into are [matplotlib style sheets](https://matplotlib.org/users/style_sheets.html) which can be used as context managers. Not quite as low-overhead as setting and forgetting, but less likely to cross package boundaries.

Do you have anybody with time to address this? If not, I can submit a PR.

effigies:
Building the docs locally, I find this doesn't change the appearance of the outputs [here](https://nistats.github.io/auto_examples/04_low_level_functions/plot_design_matrix.html). A random survey of other pages also failed to reveal changes in images.

It's unclear to me if this should be replaced with anything, or was perhaps a hack to correct some other source of global state change during development. Please let me know if you think anything more is required.

Closes #164.

### DIFF
--- a/nistats/reporting.py
+++ b/nistats/reporting.py
@@ -13,7 +13,6 @@ import matplotlib
 import matplotlib.pyplot as plt
 from patsy import DesignInfo
 from .design_matrix import check_design_matrix
-matplotlib.rc('xtick', labelsize=20)
 
 
 def compare_niimgs(ref_imgs, src_imgs, masker, plot_hist=True, log=True,


bthirion:
I don't mind this change. This line was probably copy-papsted from previous custom code.
@MartinPerez do you agree ?

effigies:
For the purposes of not depending on a pin to a repository commit, it would be handy to have this in a minor release. It's not a big deal, though.

MartinPerez:
LGTM. There should be no issues with this change. Most likely it was a quick aesthetic hack copied from somewhere else as @bthirion suggests.

Issue #166, opened 2018-02-13
Proposed changes to thresholding.get_clusters_table


tsalo:
There are a few changes I would like to propose for the `get_clusters_table` function:
1. Should the function be moved from `nistats.thresholding` to `nistats.reporting`?
2. I think it would be good to report cluster size in mm3, rather that voxels.
3. It might also be worth it to drop the matrix indices from the table, since I don’t think they’re generally very useful except for debugging.
4. When using cluster-extent thresholding (i.e., `cluster_threshold` is not None), should the table return average value, rather than peak value?
5. Should users be able to input connectivity parameters (face, edge, corner) for defining the clusters?
6. Finally, would it be useful to output subpeaks?

I have some code that does some of the things I mentioned, so I was thinking of opening a PR to merge those edits in, if they seem useful.

tsalo:
Awesome, thank you.

4. Makes sense.
5. I was thinking of adding an argument, e.g. `conn`, with accepted values of `'faces'`, `'edges'`, and `'corners'`. Alternatively, the possible values could be `{6, 18, 26}` or `{'NN1', 'NN2', 'NN3'}` (as in `3dClustSim`). Scipy's [label](https://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.ndimage.measurements.label.html) function takes in a connectivity structure that deals with this. In this case, that's just a 3x3x3 array of 1s and 0s that can be made for each of the connectivity values. I'm not sure what you mean by using a mesh, sorry.
6. SPM's tables give you up to 3 subpeaks at least 8mm apart (when available), and my code uses those parameters as well. It wouldn't be hard to adjust those two parameters though.

bthirion:
5. then I would use an argument like `neighborhood=6` (could be 18 or 26). Note that we implemented this about 10 years ago in the nipy library, but realized later that there was little usefulness for this. In practice, there's no reason not to use 6-connectivity, and I'm worried of adding complexity for no benefit.

Regarding the mesh thing: well, you might want to do this kind of thing on values sampled on a surface (cortical mesh): then you need to specifiy the mesh triangles to have the concept of local maxima. Supporting this type of data  is possible, but it's a major addition to the API.

6. Yes, that's what I had in mind. PR welcome, then.

tsalo:
I removed the input argument for connectivity from `get_clusters_table` in my fork, and I think it's about ready to for review, so I've opened a PR (#167).

[...]

PR #167, opened 2018-02-15
Changes to get_clusters_table

tsalo:
Addresses #166.

Changes:
- Move `get_clusters_table` from `nistats.thresholding` to `nistats.reporting`.
- Drop matrix space peaks from table.
- Report cluster size in mm3 instead of voxels.
- Report top 3 subpeaks more than 8mm apart when available.
- Cluster ID column uses number for cluster and lower case letter for subpeak (e.g., `1` refers to the global maximum of the first cluster, while `1a` refers to the first subpeak).
- Use voxel connectivity neighborhood of 6 by default.
- Update relevant tests.
- Add [skimage>=0.13.0](http://scikit-image.org/docs/0.13.x/) to requirements.
- Report [center of mass](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.center_of_mass.html#scipy.ndimage.center_of_mass) for clusters with >1 voxel and only one value (e.g., labels or masks), rather than peaks/subpeaks.


bthirion:
You need to configure the CI system to ass the dependency on scikit image.

bthirion:
Thinking more of it: adding an important dependency for a really minor function of the library does not seem to be a good idea. I would frankly suggest to drop this part of the PR.
@GaelVaroquaux 

To make myself clear: handling 18- or 26-connectivity is not a request from users, and could actually create unnecessary complexity. I had not realized that this would result in having skimage as a dependency.

tsalo:
I can remove the `skimage`, although it's used for extracting the cluster sub-peaks rather than labeling the clusters. The connectivity is used to label the clusters with `scipy.ndimage.measurements.label`, so there are no additional dependencies needed to account for that.

bthirion:
Sorry for my mistake.
I thinks that it is possible to extract local maxima with `scipy.ndimage`, although this is less straightforward.

tsalo:
I hadn't realized that `scipy.ndimage` had any functions that could do that, but based on [this gist](https://gist.github.com/agrimaldi/781740), it looks like you're totally right. I'll switch the code over from using `skimage` to using `scipy`, and, if it works, that will eliminate the `skimage` dependency. If I can't get it to work, I'll just drop the subpeaks from the function.

bthirion:
Agreed, thx.

tsalo:
@bthirion I'm not really sure why CircleCI is failing (something about `numpydoc`), but I think I have the `scipy`-based subpeak identification working properly. It takes a little longer than `skimage` did, but I don't think that's much of an issue, since one would probably only run `get_clusters_table` once per analysis.

KamalakerDadi:
>I'm not really sure why CircleCI is failing (something about numpydoc)

I am trying to fix that in this PR #169

nistats/reporting.py
```diff
@@ -6,16 +6,193 @@
...
+    data : array_like
+        3D array of with masked values for cluster.
+
+    min_distance : :obj:`int`
+        Minimum distance between local maxima in ``data``, in terms of
+        voxels (not mm).
+
+    Returns
+    -------
+    xy : :obj:`numpy.ndarray`
```
bthirion: 'xyz' would make more sense.

nistats/reporting.py
```diff
@@ -6,16 +6,193 @@
...
+        (n_foci, 3) array of local maxima indices for cluster.
+    """
+    data_max = ndimage.filters.maximum_filter(data, min_distance)
+    maxima = (data == data_max)
+    data_min = ndimage.filters.minimum_filter(data, min_distance)
+    diff = ((data_max - data_min) > 0)
+    maxima[diff == 0] = 0
+
+    labeled, num_objects = ndimage.label(maxima)
+    xy = np.array(ndimage.center_of_mass(data, labeled, range(1, num_objects+1)))
```
bthirion: I'd rather return peak coordinates than center of mass.

nistats/reporting.py
```diff
@@ -6,16 +6,193 @@
...
+    xy = np.round(xy).astype(int)
+    return xy
+
+
+def _get_conn(neighborhood=6):
```
bthirion: I'm afraid I don't see a good use case for that.

nistats/reporting.py
```diff
@@ -6,16 +6,193 @@
...
+
+        # Only report peak and, at most, top 3 subpeaks.
+        n_subpeaks = np.min((len(subpeak_vals), 4))
+        for subpeak in range(n_subpeaks):
+            if subpeak == 0:
+                row = [c_id+1, subpeak_xyz[subpeak, 0], subpeak_xyz[subpeak, 1],
+                       subpeak_xyz[subpeak, 2], subpeak_vals[subpeak], cluster_size_mm]
+            else:
+                # Subpeak naming convention is cluster num + letter (1a, 1b, etc.)
+                sp_id = '{0}{1}'.format(c_id+1, ascii_lowercase[subpeak-1])
```
bthirion: I think that pep8 requires spaces around the operator.

nistats/reporting.py
```diff
@@ -6,16 +6,193 @@
...
 import matplotlib
 import matplotlib.pyplot as plt
+from scipy import ndimage
 from patsy import DesignInfo
+
 from .design_matrix import check_design_matrix
 matplotlib.rc('xtick', labelsize=20)
 
 
+def _local_max(data, min_distance):
```
bthirion: This function needs a test.

nistats/reporting.py
tsalo: You're right, `xy` is not good. I think I took it from the StackOverflow answer I used as the basis for the function. Is`ijk` alright though? I'd prefer to more clearly distinguish coordinates in mm (normally `xyz`) and matrix indices.

nistats/reporting.py
tsalo: This is only for neighboring peak voxels with the same value. One thing I noticed with the tests for this function were that some of them used constant blocks of values and the old function would just return the first index, which generally corresponded so a voxel in the corner of the "cluster".

nistats/reporting.py
tsalo: Sorry about that. Added now.

tsalo:
I removed `_get_conn` and just put the 6-connectivity matrix in directly. I also had to adjust how `_local_max` was selecting foci based on distance between subpeaks (while working tests, so thanks for catching that!), but it is now working.

bthirion:
Thx, we're getting close !

bthirion:
There's very little work to finish the PR. Can you do it ? Thx.

tsalo:
@bthirion So sorry about that. I somehow missed your review. I can start working on it now.

bthirion:
That would be great, thx.

nistats/reporting.py
```diff
@@ -6,15 +6,170 @@
...
+    """Find all local maxima of the array, separated by at least min_distance.
+    Adapted from https://stackoverflow.com/a/22631583/2589328
+
+    Parameters
+    ----------
+    data : array_like
+        3D array of with masked values for cluster.
+
+    min_distance : :obj:`int`
```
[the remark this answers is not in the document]
tsalo: I'm switching to mm and including the affine. I don't think it's efficient to use an image (because it would mean generating one for each cluster in get_clusters_table), but it might increase readability. Do you have a preference?

nistats/reporting.py
```diff
@@ -6,15 +6,170 @@
...
+    maxima = (data == data_max)
+    data_min = ndimage.filters.minimum_filter(data, min_distance)
+    diff = ((data_max - data_min) > 0)
+    maxima[diff == 0] = 0
+
+    labeled, num_objects = ndimage.label(maxima)
+    ijk = np.array(ndimage.center_of_mass(data, labeled, range(1, num_objects+1)))
+    ijk = np.round(ijk).astype(int)
+
+    vals = np.apply_along_axis(arr=ijk, axis=1, func1d=_get_val, input_arr=data)
```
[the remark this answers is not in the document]
tsalo: That's what it's meant to do, but for each row in ijk since ijk is 2d. I didn't think there was a better way of doing this, although it's always struck me as a little odd.

nistats/reporting.py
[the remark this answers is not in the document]
bthirion: Passing the affine is fine imho.

tsalo:
It's not clear why circleci failed. Could it be that there was a connectivity issue during testing? The traceback points to a couple of errors like these:

```
/home/ubuntu/nistats/examples/03_second_level_models/plot_second_level_button_press.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/site-packages/nilearn/datasets/utils.py", line 565, in _fetch_file
    data = url_opener.open(request)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/urllib/request.py", line 472, in open
    response = meth(req, response)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/urllib/request.py", line 582, in http_response
    'http', request, response, code, msg, hdrs)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/urllib/request.py", line 510, in error
    return self._call_chain(*args)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/urllib/request.py", line 444, in _call_chain
    result = func(*args)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/urllib/request.py", line 590, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 503: Service Temporarily Unavailable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/nistats/doc/sphinxext/sphinx_gallery/gen_rst.py", line 450, in execute_code_block
    exec(code_block, example_globals)
  File "<string>", line 3, in <module>
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/site-packages/nilearn/datasets/func.py", line 985, in fetch_localizer_contrasts
    files = _fetch_files(data_dir, filenames, verbose=verbose)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/site-packages/nilearn/datasets/utils.py", line 738, in _fetch_files
    overwrite=overwrite)
  File "/home/ubuntu/miniconda2/envs/testenv/lib/python3.5/site-packages/nilearn/datasets/utils.py", line 585, in _fetch_file
    str(e.reason), file_name))
AttributeError: can't set attribute
```

bthirion:
Regarding the CirculeCi failure, I think that there was a problem with dataset fetching. Simply relaunch it (I could not do it.).
Best

tsalo:
@bthirion Looks like it's passing now. Thanks!

PR #169, opened 2018-02-24
FIX: CircleCI failure related to Sphinx 1.7

KamalakerDadi:
Attempts to FIX Sphinx 1.7 related issues with CircleCI. As we are seeing in this PR #167

### DIFF
--- a/doc/sphinxext/numpydoc/numpydoc.py
+++ b/doc/sphinxext/numpydoc/numpydoc.py
@@ -26,7 +26,6 @@ if sphinx.__version__ < '1.0.1':
     raise RuntimeError("Sphinx 1.0.1 or newer is required")
 
 from .docscrape_sphinx import get_doc_object, SphinxDocString
-from sphinx.util.compat import Directive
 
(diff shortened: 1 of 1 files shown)


bthirion:
Seems to work. Can you add a pedagogical comment ()why this was done ?

KamalakerDadi:
Directive is imported but not used anywhere. Moreover, module called util.compat is removed from Sphinx 1.7 that's the reason why the CircleCI is failing in other PRs.

KamalakerDadi:
Should I add comment at removed place ? I don't understand if we really want to add a comment.

bthirion:
Indeed, sorry.

effigies:
Just FYI, `numpydoc` is now its own dependency you can include rather than bundle an old version of. See, e.g. https://github.com/nipy/nipype/pull/2437.

KamalakerDadi:
@effigies Thanks. Good one! 

At some point, as a dependency of sphinx-gallery. We should upgrade to new version of sphinx-gallery which then I think would simply things as you pointed out.

PR #183, opened 2018-06-04
ENH: Return dictionary of contrast images with "all" argument

### LINKED ISSUE #180: Return multiple outputs from compute_contrast
In the first-level model, [`compute_contrast`](https://github.com/nistats/nistats/blob/master/nistats/first_level_model.py#L487) takes an `output_type` argument to indicate which map should be returned. It looks like this can only be one of several predetermined strings, which means if one wants to get multiple outputs (e.g., `effect_size` and `z`), the method has to be called multiple times. This seems both inefficient (computation doesn't appear to be cached as far as I can tell) and unwieldy. Is there any reason not to return the entire `Contrast` object and let the user pick out what they want from it? If backward compatibility is desirable, maybe `output_type` could take another value `all` or `Contrast`, in which case the actual `Contrast` is returned?

Assuming the latter is agreeable, we can work up a PR for this (looping in @effigies and @adelavega).

effigies:
This PR adds an `output_type='all'` option to `FirstLevelModel.compute_contrasts`, which causes the method to return a dictionary with each image, keyed by output type.

Closes #180.


nistats/first_level_model.py
```diff
@@ -538,26 +538,31 @@ def compute_contrast(self, contrast_def, stat_type=None,
...
-            if output_type not in ['z_score', 'stat', 'p_value', 'effect_size',
-                                   'effect_variance']:
-                raise ValueError('output_type must be one of "z_score", "stat",'
-                                 ' "p_value","effect_size" or "effect_variance"')
-        else:
-            raise ValueError('output_type must be one of "z_score", "stat",'
-                             ' "p_value","effect_size" or "effect_variance"')
+
+        valid_types = ['z_score', 'stat', 'p_value', 'effect_size',
+                       'effect_variance']
```
kchawla-pi: Why not add 'all' to the list here instead of doing a + ['all'] later?

kchawla-pi: Why not add `'all'` to the list here instead of doing a `+ ['all']` later?

nistats/first_level_model.py
```diff
@@ -538,26 +538,31 @@ def compute_contrast(self, contrast_def, stat_type=None,
...
+        if not (isinstance(output_type, _basestring) and
+                output_type in valid_types + ['all']):
```
kchawla-pi: Why do we need to check if output_type is an instance of _basestring? Seems to me simply checking that output_type is in the list of valid types will be sufficient.

nistats/first_level_model.py
```diff
@@ -538,26 +538,31 @@ def compute_contrast(self, contrast_def, stat_type=None,
...
+            raise ValueError('output_type must be one of "z_score", "stat", '
+                             '"p_value", "effect_size", "effect_variance" or '
+                             '"all"')
```
kchawla-pi: Use `str.format()` instead.

`ValueError('output_type must be one of {}'.format(valid_types))`

nistats/first_level_model.py
```diff
@@ -538,26 +538,31 @@ def compute_contrast(self, contrast_def, stat_type=None,
...
-        estimate_ = getattr(contrast, output_type)()
-        # Prepare the returned images
-        output = self.masker_.inverse_transform(estimate_)
-        contrast_name = str(con_vals)
-        output.header['descrip'] = (
-            '%s of contrast %s' % (output_type, contrast_name))
+        output_types = valid_types if output_type == 'all' else [output_type]
+
+        outputs = {}
+        for otype in output_types:
```
kchawla-pi: Make it `output_type_` instead of `otype`. That is a commonly used convention for singular interators and plural iterablas.

kchawla-pi:
Hey @effigies Thanks for this and apologies for leaving this in limbo for so long.
Would you do a pull from the master to update the TravisCI configs?

nistats/first_level_model.py
effigies: Mostly because when the actual value is `'all'`, we want access to just this list of 5 things.

```Python
output_types = valid_types if output_type == 'all' else [output_type]
```

We could set `valid_types = [..., 'all']` and then say `output_types = valid_types[:-1] if ...`, but it's not clearly improved. Whichever you prefer is fine with me.

kchawla-pi:
Thanks!

kchawla-pi:
We should add this in the What's New.rst and then I will merge it.

nistats/first_level_model.py
```diff
@@ -515,12 +515,12 @@ def compute_contrast(self, contrast_def, stat_type=None,
...
         output_type : str, optional
             Type of the output map. Can be 'z_score', 'stat', 'p_value',
-            'effect_size' or 'effect_variance'
+            'effect_size', 'effect_variance' or 'all'
 
         Returns
         -------
-        output : Nifti1Image
-            The desired output image
+        output : Nifti1Image or dict
```
bthirion: This is the only thing I dislike in this PR: that the output type depends on a rather obscure way on the input.
Yet I cannot think of any way to avoid this.
I think I would at least state here that the output type is conditional on `output_type` being `all`

kchawla-pi:
@bthirion Review & Finalize but don't do any merges yet. I am exploring some merges on Nistats and if we should revert them and I don't want to complicate it if it comes to that.

kchawla-pi:
This is good to merge and I will do so shortly.

Issue #191, opened 2018-07-20
Example of a second-level analysis on scalar (e.g. T1) images


demianw:
Hey! I think it would be informative to have a simple example performing a second-level analysis on scalar images not issued from fMRI. It will provide potential users with an easier-access tutorial for more general analyses which are not uncommon in other sub-fields of neuroimaging.

Do you think it's worth it to have one like this?

bthirion:
Yes. I'd suggest to start with OASIS, unless you have a better suggestion.

demianw:
Sounds good with me. How do we go about implementing this?

bthirion:
We write a script that does the job and then we improve it. Do you have time this week or next  Monday/Tuesday to work on it ?

demianw:
Oops missed this one. Next monday/tuesday should work for me.

bthirion:
Sorry, I'm in vacation till Aug. 13th --with bad internet connection. Would Aug. 13th work for you ?

demianw:
Yup!

D
--
Demian Wassermann, PhD
demian.wassermann@inria.fr
Associate Research Professor (CRN)
Parietal Team
INRIA Saclay Ile-de-France
1 Rue Honoré d'Estienne d'Orves, 91120 Palaiseau

PR #194, opened 2018-08-13
WIP Added Oasis example

bthirion:
This example is meant to show that it is relatively easy to run a second-level analysis with nistats, here in the case of VBM data.
Addresses #191

### DIFF
--- a/examples/03_second_level_models/plot_oasis.py
+++ b/examples/03_second_level_models/plot_oasis.py
@@ -64,8 +64,8 @@ mask_img = resample_to_img(
 from nistats.second_level_model import SecondLevelModel
 import pandas as pd
 
-design_matrix = pd.DataFrame(np.vstack((age, sex, 
-                                        np.ones(n_subjects))).T,
+intercept = np.ones(n_subjects)
+design_matrix = pd.DataFrame(np.vstack((age, sex, intercept)).T,
                              columns=['age', 'sex', 'intercept'])
 # plot the design matrix
 from nistats.reporting import plot_design_matrix
@@ -74,7 +74,9 @@ ax.set_title('Second level design matrix', fontsize=12)
 ax.set_ylabel('maps')
 plt.tight_layout()
 
+##########################################################################
 # specify and fit the model
+
 second_level_model = SecondLevelModel(smoothing_fwhm=2.0, mask=mask_img)
 second_level_model.fit(gray_matter_map_filenames,
                        design_matrix=design_matrix)
@@ -94,7 +96,7 @@ _, threshold = map_threshold(
 
 display = plotting.plot_stat_map(
     z_map, threshold=threshold, colorbar=True, display_mode='z',
-    cut_coords=3,
+    cut_coords=[-4, 26],
     title='age effect on grey matter density (FDR < .05)')
 
 ###########################################################################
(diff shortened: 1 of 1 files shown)


bthirion:
@demianw please take a look and comment.

examples/03_second_level_models/plot_oasis.py
```diff
@@ -0,0 +1,115 @@
...
+from nilearn import datasets
+from nilearn.input_data import NiftiMasker
+
+n_subjects = 100  # more subjects requires more memory
+
+############################################################################
+# Load Oasis dataset
+# -------------------
+oasis_dataset = datasets.fetch_oasis_vbm(n_subjects=n_subjects)
+gray_matter_map_filenames = oasis_dataset.gray_matter_maps
```
demianw: @bthirion I think that a pedagogic issue with this is that it's unclear how to assemble a dataset to be analysed by `nistats`. 

* Do I always need to save all the images on my dataset and then give it a list of filenames?
* Can I give it a 4-dimensional image? If yes, over what dimension are we performing the analysis?
* Can I give it a list of images?

bthirion: You can give either a 4-dimensional image or a list of 3D images.
You can give either an already-loaded  NiftiImage(s) or a path to such image(s).
I can't figure out any difficulty here.

demianw: This doesn't seem to work
```
from glob import glob
import nibabel as nib
import pandas
from nistats.second_level_model import SecondLevelModel
import numpy as np

fs = glob('*.nii.gz') # Assuming all images have the same dimensions

data = nib.spatialimages.SpatialImage(
    np.concatenate([nib.load(f).get_data()[..., None] for f in fs], axis=-1),
    nib.load(fs[0]).affine
)
second_level_model = SecondLevelModel(smoothing_fwhm=2.0)

c = pandas.DataFrame([[1]] * len(fs), columns=['intercept'])

r1 = second_level_model.fit(ddata, design_matrix=c)
```

Failing with
`ValueError: second_level_input must be a list of 'FirstLevelModel' objects, a pandas DataFrame or a list Niimg-like objects. Instead <class 'nibabel.nifti1.Nifti1Image'> was provided`

Also tried
```
data = nib.spatialimages.Nifti1Image(
    np.concatenate([nib.load(f).get_data()[..., None, None] for f in fs], axis=-1),
    nib.load(fs[0]).affine
)
```
which is more inline with [nifti specs](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h) (4-th axis is temporal and 5th vector-valued data)

bthirion: `second_level_model.fit(fs, design_matrix=c)` works.
However, I agree that in the current state, the code is not flexible enough to accept a 4D image.

demianw: Yes! There is a way to make it work. Its just that the 4D image is a common application case. Either turning it onto a list of images or having to save each image separately seems not too ergonomic to me.

bthirion: Besides  4D image aspect that will be handled in another PR, can we consider merging this PR ?

bthirion:
Could we merge that one and proceed with the 4D image pending issue (#195) with PR #214.
If nobody objects, I'll merge tomorrow.

PR #196, opened 2018-08-24
[WIP] Pedagogical example of surface-based analysis

bthirion:
Still needs some more explanation, but the code is relatively simple.


bthirion:
Merging.

Issue #201, opened 2018-08-27
Second level model should have necessarily an intercept


bthirion:
Systematically enforcing an intercept will lead to rank-deficient model in some situations, which is bad.
Related to #162 

proposition:
1. remove the obligation to have an intercept
2. add an example showing a two-sample t-test with 2 * n images, i.e. 2 conditions * n subjects, with a modeling of the subject effect while testing for the condition effect.

PR #203, opened 2018-08-28
Intercept optional in second level model

bthirion:
Addresses #201


bthirion:
@effigies @chrisfilo
Does this address your concern -- at least in a rudimentary way ?

bthirion:
Merging.

Issue #182, opened 2018-05-31
Bug: Oversampling during convolution with a high temporal resolution regressor


adelavega:
Given a regressor with high temporal frequency (in this example, speech word onset in a movie), it seems that the re-sampling ("oversampling") produces strange results in `make_design_matrix`.

For example, this is my regressor (this is the output of `check_paradigm`). Onsets are sometimes only a few milliseconds apart, and durations are similarly short. (onset is in seconds, TR=1.5)
![image](https://user-images.githubusercontent.com/2774448/40755548-75a8e9c4-6444-11e8-94b2-10ecf7e5cfbe.png). 

Given the default oversampling of 16, the following regressor is created:

     import nistats.hemodynamic_models as hm
     convolved = hm.compute_regressor(exp_condition, 'glover', frame_times, oversampling=16)[0]
     pd.DataFrame(convolved).plot()

![image](https://user-images.githubusercontent.com/2774448/40755722-31692660-6445-11e8-9e12-59a8b4a27c3b.png)

Raising the oversampling to a higher value (e.g. 50) produces a more sensible result:

![image](https://user-images.githubusercontent.com/2774448/40755717-290e2056-6445-11e8-82d2-50d6e9e0b636.png)

Perhaps, allowing users to access the oversampling parameter from the `make_design_matrix` API would be a good idea, but ideally the default works. This doesn't seem to be a problem with lower resolution events (e.g. every 1-1.5s)

bthirion:
Can you issue a PR ?

bthirion:
Addressed in PR #209

bthirion:
Closing wit merge of #209

PR #209, opened 2018-09-05
[MRG] exposed oversampling parameter

bthirion:
Addresses #182 
Please review

### DIFF
--- a/nistats/design_matrix.py
+++ b/nistats/design_matrix.py
@@ -164,7 +164,7 @@ def _make_drift(drift_model, frame_times, order=1, period_cut=128.):
 
 
 def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
-                         min_onset=-24):
+                         min_onset=-24, oversampling=None):
     """ Creation of  a matrix that comprises
     the convolution of the conditions onset with a certain hrf model
 
@@ -191,6 +191,10 @@ def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
         Minimal onset relative to frame_times[0] (in seconds) events
         that start before frame_times[0] + min_onset are not considered.
 
+    oversampling: float or None, optional,
+        Oversampling factor used in temporal convolutions. Should be 1 for
+        whenever hrf_mode is 'fir' and 16 otherwise.
+
     Returns
     -------
     regressor_matrix : array of shape (n_scans, n_regressors),
@@ -210,8 +214,11 @@ def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
     regressor_names = []
     regressor_matrix = None
     if hrf_model == 'fir':
+        if oversampling not in [1, None]:
+            warn('Forcing oversampling factor to 1 for a finite'
+                 'impulse response hrf model')
         oversampling = 1
-    else:
+    elif oversampling is None:
         oversampling = 16
 
     trial_type, onset, duration, modulation = check_paradigm(paradigm)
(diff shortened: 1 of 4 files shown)


bthirion:
Any review on that one ?

KamalakerDadi:
Issue #182 hints out at the example "An example showcasing this behavior would actually be useful."

Any plans for an example.

KamalakerDadi:
Can you add changes into doc/whats_new.rst ?

nistats/design_matrix.py
```diff
@@ -274,7 +281,7 @@ def _full_rank(X, cmax=1e15):
 def make_design_matrix(
     frame_times, paradigm=None, hrf_model='glover',
     drift_model='cosine', period_cut=128, drift_order=1, fir_delays=[0],
-    add_regs=None, add_reg_names=None, min_onset=-24):
+    add_regs=None, add_reg_names=None, min_onset=-24, oversampling=None):
```
KamalakerDadi: In some functions oversampling is by default 16. Is there any reason to keep default to None and then convert to 16.

nistats/design_matrix.py
```diff
@@ -332,6 +339,10 @@ def make_design_matrix(
         Minimal onset relative to frame_times[0] (in seconds)
         events that start before frame_times[0] + min_onset are not considered.
 
+    oversampling: float or None, optional,
+        Oversampling factor used in temporal convolutions. Should be 1 for
+        whenever hrf_mode is 'fir' and 16 otherwise.
```
KamalakerDadi: Typo here should be hrf_model

nistats/design_matrix.py
```diff
@@ -191,6 +191,10 @@ def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
         Minimal onset relative to frame_times[0] (in seconds) events
         that start before frame_times[0] + min_onset are not considered.
 
+    oversampling: float or None, optional,
+        Oversampling factor used in temporal convolutions. Should be 1 for
+        whenever hrf_mode is 'fir' and 16 otherwise.
```
KamalakerDadi: Same here typo hrf_model

nistats/design_matrix.py
```diff
@@ -210,8 +214,11 @@ def _convolve_regressors(paradigm, hrf_model, frame_times, fir_delays=[0],
     regressor_names = []
     regressor_matrix = None
     if hrf_model == 'fir':
+        if oversampling not in [1, None]:
+            warn('Forcing oversampling factor to 1 for a finite'
+                 'impulse response hrf model')
```
KamalakerDadi: Can you increase the stacklevel here ? So that it won't float around other verbose messages.

KamalakerDadi:
I don't have other comments as far as I can see.

adelavega:
Looks good to me. Thanks for the PR

adelavega:
Question: Is there a reason for the default parameter being 16? As I pointed out in my original issue, this is too low for high-frequency effects, and leads to odd behavior. Is there a downside to making the default a higher value? Or would this have unintended consequences for lower frequency events?

bthirion:
I think I picked 16 because this is the SPM default. Making it larger would result in additional computational cost, but it's negligible. The true point is: in 99% of experimental designs people don't use very high frequency signals, as these fast changes will be killed by hrf convolution anyhow...
What deafult value do you advocate ? 32 ?

adelavega:
We've been using 50, and that seems to work for natural speech (word speech rate). Personally, I'm happy to manipulate this parameter as I need, but I think a default that covers a larger set of use cases is always good (given negligible increased computation, relative to model fitting)

bthirion:
"An example showcasing this behavior would actually be useful."
Yes, I wrote that. I'm currently refactoring the doc, and I think that exposing this parameter  is not a high priority, and I propose to drop it in the short term.

bthirion:
More comments on that one ?

PR #210, opened 2018-09-05
[WIP] T fcontarst cleaning

bthirion:
This PR is a massive code simplification. We should decide whether we want it or not.


nistats/model.py
```diff
@@ -45,7 +46,7 @@ def __init__(self, theta, Y, model, cov=None, dispersion=1., nuisance=None,
             multiplicative factor in front of `cov`
 
         nuisance : None of ndarray
-            parameter estimates needed to compute logL
+            parameter estimates needed to compute log-lokelohood
```
kchawla-pi: Typos in "lokelohood"

kchawla-pi:
If you are removing unused or obviateable code then that's good.
What's the uncertainty in removing these functions?

kchawla-pi:
What's the status on this?

bthirion:
This is meant to remove a large part of the code base that is not great (over-designed). I can try and revive it.

kchawla-pi:
Indeed. This seems like an important one before we incorporate this into Nilearn.

kchawla-pi:
@bthirion I think if you can complete Jerome's PR #416 that will serve the project better than completing this one.

Issue #189, opened 2018-07-20
"3.4.2. GLM fitting in second level fMRI" example only shows a trivial second-level model


bthirion:
Find a dataset in which there are putative population effects.

bthirion:
Addressed, closing.

PR #212, opened 2018-09-05
[WIP] Added another second-level example

bthirion:
Address #189 
Should probably be integrated with #203


bthirion:
Will merge soon unless I hear otherwise.

Issue #223, opened 2018-09-14
Should accpet .tsv files as onset files


bthirion:
Currently, the library assumes a csv format. Should actually use tsv. At least in a near future, should support both.
(PR #228 - KC)

kchawla-pi:
Does it not already?
`pandas.from_csv()` seems to automatically figures out whether the separators/delimiters are tabs or commas and creates a pandas dataframe regardless. 
The functions in `nistats.first_level_model.py` have been coded to expect tsv, not csv. (they all use `pandas.from_csv(sep='\t')` not the default `sep=','` .
That csv is expected seems to be mentioned only in the documentation, not what is actually happening. Is that what needs fixing? Only the documentation?

bthirion:
The small example  below triggers an error, because the paradigm file is not handled correctly
```
from nistats.first_level_model import FirstLevelModel
from nistats import datasets
data = datasets.fetch_localizer_first_level()
paradigm_file = data.paradigm
fmri_img = data.epi_img
first_level_model = FirstLevelModel(2.4, 0)
first_level_model = first_level_model.fit(fmri_img, paradigm_file)
```

kchawla-pi:
The paradigm file in this example's data set is neither csv nor tsv, the values are separated by single spaces, not commas or tabs. How should we handle this?

kchawla-pi:
> we should fix the csv file in the fetcher.
That would mean fixing the file and upload it to the online database, ftp.cea.fr .
How do we do that?

bthirion:
I may find it, but this probably requires being on the CEA inrtanetwork. A hotfix is just to read and re-write the file after download ---in the fetcher function-- and return a correct file to the user.
Actually, we need to do that for all datasets fetched / imported / used in nistats. The experimental paradigm descriptors were set long before BIDS existed and should be trabsformed to regular BIDS onset files.

kchawla-pi:
@KamalakerDadi 
nilearn._utils.numpy_conversions.csv_to_array managed to correctly load that file and may do so for several others.
We still will need a function to flag improper events file instead of transparently loading them.

kchawla-pi:
@bthirion Even after the events file verify PR merge, the example you ave above still raises `ValueError: The provided paradigm has no onset key
`. I discovered this is because `utils._check_and_load_tables()` when reading the events file as a Pandas dataframe `pd.read_csv(table, index_col=0)`, is not reading the headers.

bthirion:
[...]

PR #228, opened 2018-09-19
[MRG] Raise error for BIDS non-compliant events files

kchawla-pi:
for issue #223


kchawla-pi:
Read my comment about #223  first.

kchawla-pi:
Working on pushing unit tests.

nistats/utils.py
```diff
@@ -41,6 +42,28 @@ def _check_and_load_tables(tables_, var_name):
...
+        except IOError:
+            pass
+        else:
+            try:
+                csv.Sniffer().sniff(sample=events_file_sample,
+                                    delimiters=valid_separators,
+                                    )
+            except csv.Error:
+                raise ValueError(
+                    'The values in the events file are not separated by tabs or commas',
```
bthirion: Actually I would put "The values in the events file are not separated by tabs; please enforce BIDS conventions".

kchawla-pi:
Ready for review @bthirion

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -0,0 +1,147 @@
+import csv
```
bthirion: should be  nistats/tests/test_verify_events_file_uses_tab_separators.py
i don't see the usefulness of adding the unit_tests levels as all our tests are unit_tests anyhow.

nistats/utils.py
```diff
@@ -40,6 +41,67 @@ def _check_and_load_tables(tables_, var_name):
...
+    this may change in future as commas are not BIDS compliant.
+    
+    parameters
+    ----------
+    events_files: str, List/Tuple[str]
+        A single file's path or a collection of filepaths.
+        Files are expected to be text files.
+        Non-text files will raise ValueError.
+    
+    Returns
```
bthirion: Does it it need to return anything ?

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -0,0 +1,147 @@
...
+    temp_data_bytearray_from_ints = bytearray([0, 1, 0, 11, 10])
+    with NamedTemporaryFile(mode='wb', dir=os.getcwd(),
+                            suffix='.bin') as temp_bin_obj:
+        temp_bin_obj.write(temp_data_bytearray_from_ints)
+        with assert_raises(ValueError):
+            result = _verify_events_file_uses_tab_separators(
+                    events_files=temp_bin_obj.name)
+
+
+if __name__ == '__main__':
```
bthirion: I'm not familiar with pytests, but is this really necessary ? normally, the tests are run even without the main file ...

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
kchawla-pi: I have no problem in putting them in tests.

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
kchawla-pi: This enables me to run the tests without invoking the frameworks, which is useful when I want to debug the tests, since I can set regular breakpoints like in standard python (not having to run it as a framework).
It also helps with debugging if something breaks in future . Running them only using the testing framework precludes this ability.
I can remove it, once the tests have been written successfully, or I can add a comment explaining its purpose. Either way it won't interfere with the code and is not difficult to understand.
What would you like?

nistats/utils.py
kchawla-pi: Not for functionality and I personally don't find this approach very elegant. I wanted to make sure I had a way to test the do-nothing cases and this was the most straightforward way. If I really wanted to, I suppose I should implement the logging module maybe.
If I do remove this, I will also remove the unit tests that depend on this. There's a consideration that those unit tests may not be necessary at all.
What do you think?

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
kchawla-pi: There's also the idea that I will be adding more unit tests as time goes on, and the current tests are more integration tests than unit tests- they don't test individual functions or methods but small pipeline of multiple functions.
I will defer to your opinion here. Let me know.

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
bthirion: Makes sense, but I think that we should organize tests uniformally across modules (and is ppossible, with Nilearn).
For the time being, we can keep it as is -- it certainly does not harm.

nistats/tests/unit_tests/test_verify_events_file_uses_tab_separators.py
bthirion: Maybe, but for the moment I'd like to keep the thing homogeneous (and homogeneous with Nilearn). So let's keep that on a TODO list.

nistats/utils.py
bthirion: For the tests, cant'you simply use assert_raises ?

bthirion: I think I get your point. However, I'm worried about API homogeneity: I would not like to introduce custom function signatures, simply to ensure that a unit test specifically tests one function. 
@GaelVaroquaux  any opinion on this ?

kchawla-pi: I agree, this is rather inelegant; if there is a better way to do this I would like to know.
We should start introducing per function testing now, it will improve the code and coding practices by future contributors.

kchawla-pi: I will remove the section returning the exceptions and modify the tests accordingly.
We should start insisting on er function unit tests from new contributors & contributions.

bthirion:
Thx.

nistats/tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -0,0 +1,137 @@
...
+    temp_csv.flush()
+
+
+def _run_test_for_invalid_separator(filepath, delimiter_name):
+    if delimiter_name not in ('tab', 'comma'):
+        with assert_raises(ValueError):
+            _verify_events_file_uses_tab_separators(events_files=filepath)
+    else:
+        result = _verify_events_file_uses_tab_separators(events_files=filepath)
+        assert_true(result is None)
```
bthirion: Then this should always be the case.

bthirion:
Almost ready to merge, thx !

nistats/tests/test_verify_events_file_uses_tab_separators.py
bthirion: Sorry. 
results is alway None, whether the file is valid or None.

kchawla-pi: The verification function raises a ValueError if the file is a text file with valid path BUT does not use tabs or commas.
Every other case, the verification function returns None. Other errors are caught by other functions in the original code.
Here the comment is about the unit test I wrote for the verification function.

bthirion: All right, you're too subtle for me.

bthirion:
This one looks ready for merge. I'll do so tonight unless somebody jumps in.

bthirion:
OK, merging.

PR #236, opened 2018-09-27
Fix wrong paradigm key check

rschmaelzle:
I believe that the way the keys are checked is wrong... (this is my first github-pull-request - so sorry if I did it wrong  - you may want a different way to fix this ... but I thought I'd make a note since I ran into it...)


bthirion:
Indeed!

But the logic should be the opposite: remain silent if this is all right, and  raise an Error when the conventions are broken and a warning if the events descriptor is not well-structures. So, I just took a look at BIDS specification and it turns out that onset and duration are required. Hence
* raise an error of onset id not provided
* raise an error if duration is not provided
* raise a warning if trial_type is not povided (I think it should)
* print a comment if `modulation` is found as this is currently used to 
and that's all.
Can you do that ?

rschmaelzle:
i tried to - need get more savvy with github and nistats (didn't understand all re: modulation).

bthirion:
Do you need help on that one ?
FYI we're going to release early next week, and we absolutely need this improvement in. If you don't have time to work on it, I can contribute the missing parts.

rschmaelzle:
@bthirion I will try to understand and do it tonight and tomorrow, but I can't guarantee since I am still a newbie. If you want to be sure, you can do it. Otherwise, I'll sent sth by tomorrow night.

rschmaelzle:
the checks are still failing - it seems because the fetched example design matrices also don't contain the bids-duration - e.g. fetch_localizer_first_level. I may fix those as well - but only if I am on the right track and not creating a mess.

bthirion:
All right, it is fixed now !
Reviews welcome. If no issue is raised we'll merge it during the week end.

nistats/experimental_paradigm.py
```diff
@@ -48,19 +48,18 @@ def check_paradigm(paradigm):
...
     modulation = np.ones(n_events)
-    if 'trial_type' in paradigm.keys():
+    if 'trial_type' not in paradigm.keys():
         warnings.warn("'trial_type' key not found in the given paradigm.")
-        trial_type = np.array(paradigm['trial_type'])
-    if 'duration' in paradigm.keys():
-        warnings.warn("'duration' key not found in the given paradigm.")
-        duration = np.array(paradigm['duration']).astype(np.float)
+        trial_type = np.repeat('dummy', n_events)
     if 'modulation' in paradigm.keys():
```
kchawla-pi: Do we need to generate a warning if the key is found or if the key is not found?

bthirion: when it is found, so that the user is aware that the information will be used in the design matrix computation (this only represents a minority of cases).

bthirion:
OK, merging since this is fixed.

Issue #234, opened 2018-09-23
Convert non-tsv files to tsv during dataset fetch


kchawla-pi:
See https://github.com/nistats/nistats/issues/223#issuecomment-423317417

WE need to write a function called by the dataset fetchers whenever a dataset is accessed where the events are written not as csv r tsv but with a different delimiter character, such as spaces.
Presumably this will be added to NiLearn as well.

kchawla-pi:
No I have not.Thanks, I will check this out.

bthirion:
'fetch_spm_multimodal_fmri'
```
timing = loadmat(getattr(subject_data, "trials_ses%i" % (idx + 1)),
                     squeeze_me=True, struct_as_record=False)
    faces_onsets = timing['onsets'][0].ravel()
    scrambled_onsets = timing['onsets'][1].ravel()
    onsets = np.hstack((faces_onsets, scrambled_onsets))
    onsets *= tr  # because onsets were reporting in 'scans' units
    conditions = (['faces'] * len(faces_onsets) +
                  ['scrambled'] * len(scrambled_onsets))
    paradigm = pd.DataFrame({'trial_type': conditions, 'onset': onsets})

```
(see e.g. https://nistats.github.io/auto_examples/02_first_level_models/plot_spm_multimodal_faces.html#sphx-glr-auto-examples-02-first-level-models-plot-spm-multimodal-faces-py)

In 'fetch_spm_auditory'
```
import numpy as np
tr = 7.
epoch_duration = 6 * tr 
conditions = ['rest', 'active'] * 8
n_blocks = len(conditions)
duration = epoch_duration * np.ones(n_blocks)
onset = np.linspace(0, (n_blocks - 1) * epoch_duration, n_blocks)

import pandas as pd
events = pd.DataFrame(
    {'onset': onset, 'duration': duration, 'trial_type': conditions})
```
(see e.g. https://nistats.github.io/auto_examples/01_tutorials/plot_spm_auditory.html#sphx-glr-auto-examples-01-tutorials-plot-spm-auditory-py)

kchawla-pi:
I am getting the error `connection refused` when I try to use `nistats.datasets.fetch_localizer_first_level(<new location>)` to a new location.
Anyone else seeing this? Ideas why this is happening?

PR #238, opened 2018-10-01
[MRG] Make events files fetched via dataset fetchers BIDS compliant

kchawla-pi:
issue #234 
First-level localizer fMRI dataset events file is now BIDS compliant


bthirion:
Please also consider the other diowloaded files (spm_auditory, spm_multimodal_faces).
Thx.

kchawla-pi:
Yeah, I figured I'll work first on https://github.com/nistats/nistats/issues/223#issuecomment-426573423
. You want to wrap this up first?

bthirion:
To me these things go together. You can handle them in the same PR if it's easier.

kchawla-pi:
From now on, and in all examples, while reading events files, we'll either have to use `pandas.read_csv(sep='\t')` or `pandas.read_table()` .
We should check and modify the examples accordingly.

kchawla-pi:
@bthirion 2nd fetcher landed. Ready for review.

examples/01_tutorials/plot_single_subject_single_run.py
```diff
@@ -65,38 +65,13 @@
...
-    42  ,  504  ,  rest
-    42  ,  546  ,  active
-    42  ,  588  ,  rest
-    42  ,  630  ,  active
-"""
-
-###############################################################################
-# We can read such a table from a spreadsheet file  created with OpenOffice Calcor Office Excel, and saved under the *comma separated values* format (``.csv``). 
+# timing of the auditory stimulation and rest periods. This is typically
+# provided in an events.tsv or events.csv file. The path of this file is
```
bthirion: I'd simply put: "events.tsv file"

examples/01_tutorials/plot_single_subject_single_run.py
```diff
@@ -65,38 +65,13 @@
...
+# provided in the dataset.
+# Please note that events are sometimes called paradigms, though this
```
bthirion: I'd remove that statement. I don't think we should explicitly argue about terminology

nistats/tests/test_make_bids_compliant_localizer_first_level_paradigm_file.py
```diff
@@ -0,0 +1,52 @@
+import os
```
bthirion: I'm wondering why you're not adding that one in test_datasets.py directly. I think I wouldn't like to see 100 files in this directory.

nistats/tests/test_make_spm_auditory_events_file.py
```diff
@@ -0,0 +1,44 @@
+import os
```
bthirion: Same thing here.

bthirion:
Let me know when this is ready for review.

kchawla-pi:
Absolutely. 
Currently investigating a weird glitch where the tests are failing because the .mat files are empty during nosetests, but are fine when I run the fetch_spm_multimodal_fmri() .

kchawla-pi:
It seems the failure is due to the way the test is written. Dummy .mat files are created (which are empty) during the tests and this was fine because those .mat files were not being loaded during testing.
However, when the events file are being generated in the fetcher itself, that requires the .mat files to be loaded using `loadmat()`. Since they are empty, the tests are failing.

kchawla-pi:
@bthirion 
For now I have programmed it to generate a warning during nosetests about the inability to generate events files when the Mat files are empty.
A solution would be to download the dataset from the server when this happens, instead of the mock data currently used. However that will add 2 minutes of download time everytime the tests are run locally.
I haven't written a unit test for this yet, other than that, this is ready for review.

nistats/datasets.py
```diff
@@ -334,69 +485,134 @@ def fetch_spm_auditory(data_dir=None, data_name='spm_auditory',
...
+        if not _subject_data:
+            return None
+        try:
+            paradigm = _make_events_file_spm_multimodal_fmri(_subject_data, session)
+        except MatReadError as mat_err:
+            warnings.warn('{}. An events.tsv file cannot be generated'.format(str(mat_err)))
+        else:
+            events_filepath = _make_events_filepath_spm_multimodal_fmri(_subject_data, session)
+            paradigm.to_csv(events_filepath, sep='\t', index=False)
+            _subject_data['events{}'.format(session+1)] = events_filepath
```
bthirion: pep8 'session + 1'

nistats/datasets.py
```diff
@@ -334,69 +485,134 @@ def fetch_spm_auditory(data_dir=None, data_name='spm_auditory',
...
+    return _glob_spm_multimodal_fmri_data(subject_dir)
 
-        return Bunch(**_subject_data)
 
-    # maybe data_dir already contains the data ?
-    data = _glob_spm_auditory_data()
-    if data is not None:
-        return data
+def _make_events_filepath_spm_multimodal_fmri(_subject_data, session):
+    key = 'trials_ses{}'.format(session+1)
```
bthirion: same comment

nistats/datasets.py
```diff
@@ -334,69 +485,134 @@ def fetch_spm_auditory(data_dir=None, data_name='spm_auditory',
...
+    events_file_location = os.path.dirname(_subject_data[key])
+    events_filename = 'session{}_events.tsv'.format(session+1)
```
bthirion: again

nistats/datasets.py
kchawla-pi: In standalone expressions, not in function/method args.

nistats/datasets.py
kchawla-pi: In standalone expressions, not in function/method args

nistats/datasets.py
kchawla-pi: In standalone expressions, not in function/method args

kchawla-pi: Regardless, I'll obviate this concern soon

nistats/datasets.py
kchawla-pi: Regardless, I'll obviate this concern soon

nistats/datasets.py
kchawla-pi: Regardless, I'll obviate this concern soon

kchawla-pi:
I can't write the missing unit test withut bundling a 1.2KB .mat file, hence that will keep.
One final review and this is done AFAIK.

kchawla-pi:
Make sure to update the nilearn_data using the new fetchers for the tutorial. Else the new code will not work.

bthirion:
It's fine to ship a 1.2K mat file with the library.

kchawla-pi:
Turns out it is more complex than that. I don't think we need to focus on this right away, post release, I will add tests.

bthirion:
Can you rebase on master ?

bthirion:
Great ! I'll look at that tomorrow.

bthirion:
Great job, merging.

kchawla-pi:
Cool, I'll get started on the events not paradigms PR.

Issue #221, opened 2018-09-12
thresholding API needs polishing


bthirion:
- When `fpr` is used, it should not require an image
- it should not have an argument called `threshold`

PR #242, opened 2018-10-03
[MRG] changed the API and docstring of map_threshold

bthirion:
Addresses #221

### DIFF
--- a/nistats/thresholding.py
+++ b/nistats/thresholding.py
@@ -45,9 +45,9 @@ def map_threshold(stat_img=None, mask_img=None, level=.001,
     ----------
     stat_img : Niimg-like object or None, optional
        statistical image (presumably in z scale)
-       whenever height_control is 'fpr' or 'none', 
-       stat_img=None os acceptable. 
-       If it is 'fdr' or 'bonferroni', an error is raised       
+       whenever height_control is 'fpr' or None,
+       stat_img=None is acceptable.
+       If it is 'fdr' or 'bonferroni', an error is raised if stat_img is None.
 
     mask_img : Niimg-like object, optional,
         mask image
--- a/nistats/thresholding.py
+++ b/nistats/thresholding.py
@@ -15,7 +15,7 @@ def fdr_threshold(z_vals, alpha):
     Parameters
     ----------
     z_vals: array,
-            a set of z-variates from which an FDR
+            a set of z-variates from which the FDR is computed
     alpha: float,
            desired FDR control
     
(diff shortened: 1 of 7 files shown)


bthirion:
Please review.

KamalakerDadi:
I am looking at the issue. I did not understand this point "When fpr is used, it should not require an image". Could you say more details on this in this PR ?

KamalakerDadi:
>Well, when you simply control the FPR, the threhsold does not depend on the data you entered: it only depdends on the targeted p-value. In that sense, the function should not even require an image. Does that makes sense ?

function here you mean ```map_threshold``` ?

nistats/thresholding.py
```diff
@@ -10,21 +10,36 @@
...
+        return (z_vals_[pos][-1] - 1.e-12)
     else:
         return np.infty
 
 
-def map_threshold(stat_img, mask_img=None, threshold=.001,
+def map_threshold(stat_img, mask_img=None, level=.001,
                   height_control='fpr', cluster_threshold=0):
-    """ Threshold the provided map
+    """ Compute the required threhsold level and return the thresholded map
```
KamalakerDadi: threshold

nistats/thresholding.py
```diff
@@ -34,15 +49,21 @@ def map_threshold(stat_img, mask_img=None, threshold=.001,
     mask_img : Niimg-like object, optional,
         mask image
 
-    threshold: float, optional
-        cluster forming threshold (either a p-value or z-scale value)
+    level: float, optional
+        number controling the thresholding (either a p-value or z-scale value)
+        Not te be confused with the z-scale threshold: level can be a p-values,
```
KamalakerDadi: to

nistats/thresholding.py
```diff
@@ -10,21 +10,36 @@
 
 
 def fdr_threshold(z_vals, alpha):
-    """ return the BH fdr for the input z_vals"""
+    """ return the Benjamini-Hochberg FDR threshold for the input z_vals
+    
+    Parameters
+    ----------
+    z_vals: array,
+            a set of z-variates from which an FDR
```
KamalakerDadi: I would write "a set of z-variates from which an FDR is calculated."

KamalakerDadi:
I think this changes deserves to add in change log.

KamalakerDadi:
Because there is no deprecation warning to remove threshold parameter

bthirion:
Will probably merge with in the next 24hours unless I hear otherwise.

nistats/thresholding.py
```diff
@@ -10,48 +10,98 @@
...
-    """ Threshold the provided map
+    """ Compute the required threshold level and return the thresholded map
 
     Parameters
     ----------
-    stat_img : Niimg-like object,
+    stat_img : Niimg-like object or None, optional
        statistical image (presumably in z scale)
+       whenever height_control is 'fpr' or 'none', 
+       stat_img=None os acceptable.
```
KamalakerDadi: is

KamalakerDadi: May be: "For height_control='fpr', stat_img has no effect and can be given as None."

nistats/thresholding.py
```diff
@@ -10,48 +10,98 @@
...
+    """ Compute the required threshold level and return the thresholded map
 
     Parameters
     ----------
-    stat_img : Niimg-like object,
+    stat_img : Niimg-like object or None, optional
        statistical image (presumably in z scale)
+       whenever height_control is 'fpr' or 'none', 
+       stat_img=None os acceptable. 
+       If it is 'fdr' or 'bonferroni', an error is raised
```
KamalakerDadi: an error is raised if stat_img is not provided.

nistats/thresholding.py
bthirion: It is actually a bit more subtle: when the image is provided, a thresholded version is returned. For this readons, I prefer to keep the current formulation.

bthirion:
Thx for the review ! 
Will merge that one soon.

PR #214, opened 2018-09-07
[WIP] 4D image no longer fail parameter checks in SecondLevelModel.fit()

kchawla-pi:
- Added function: validation_4d_images()
 - Added function: verify_4d_shape()
 - Now getting slice errors further down in the SecondLevelModel.fit() .

For issue #195

PR #244, opened 2018-10-04
[MRG] Ad hoc fix of 4D image problem

bthirion:
A simpler attempt to solve #195 (alternative to #214, that involves larger refactoring).

### DIFF
--- a/nistats/second_level_model.py
+++ b/nistats/second_level_model.py
@@ -19,6 +19,7 @@ from sklearn.base import BaseEstimator, TransformerMixin, clone
 from nilearn._utils.niimg_conversions import check_niimg
 from nilearn._utils import CacheMixin
 from nilearn.input_data import NiftiMasker
+from nilearn.image import mean_img
 from patsy import DesignInfo
 
 from .first_level_model import FirstLevelModel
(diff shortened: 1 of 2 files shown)


bthirion:
Will probably merge with in the next 48hours unless I hear otherwise.

nistats/second_level_model.py
```diff
@@ -262,6 +269,8 @@ def fit(self, second_level_input, confounds=None, design_matrix=None):
             sample_map = second_level_input['effects_map_path'][0]
             labels = second_level_input['subject_label']
             subjects_label = labels.values.tolist()
+        elif isinstance(second_level_input, Nifti1Image):
+            sample_map = mean_img(second_level_input)
```
KamalakerDadi: Wasn't it the idea of operating on list of 3D images given a 4D image ? But, it seems you are averaging.

bthirion: The idea was to avoid using second_level_input[0], as second_level_input is not necessarily a list. Also, IIRC the sample_map is used for masking, hence it is actually better to base this on the mean image. Does that make sense ?

bthirion:
Merging now.

PR #255, opened 2018-10-07
[MRG] Change minimum numpy ver to 1.11.0  for NiLearn compatibility

kchawla-pi:


continuous_integration/circle_ci_test_doc.sh
```diff
@@ -1,10 +0,0 @@
-#!bin/bash
```
bthirion: Why do you need to remove that ?

bthirion:
LGTM then.

kchawla-pi:
Are you still getting that error?

bthirion:
I can only check that  at NSP :-(

bthirion:
If nobody protests I'll merge within 1hour.

PR #259, opened 2018-10-08
[MRG] boto3/botocore is imported only for openneuro fetcher

### LINKED ISSUE #254: 'botocore' is necessary to use 'datasets' modules
It should ot be the case. Botocore should be required only for the functions that need it (very few actually).
The dependece on botocore should be indicated on a by-script/module basis.

kchawla-pi:
Fixes #254

### DIFF
--- a/nistats/datasets.py
+++ b/nistats/datasets.py
@@ -10,7 +10,6 @@ import os
 import re
 import warnings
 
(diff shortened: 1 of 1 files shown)


kchawla-pi:
@bthirion

bthirion:
OK, great. Merging.

PR #264, opened 2018-10-08
[MRG] Stats intro

bthirion:
Added statistical paragraphs to the introduction tutorial.
Reviews welcome, but we'll need to merge that quickly.


doc/introduction.rst
```diff
@@ -35,6 +38,9 @@ A cerebral MR image provides a 3D image of the brain that can be decomposed into
 
 .. note:: A typical step in the preprocessing of MR images, involves spatially morphing these images onto a standard template (e.g. the MNI152 template from the Montreal Neurological Institute). One this is done, the coordinates of a voxel are in the same space as the template and can be used to estimate its brain location using brain atlases. As already mentioned, the nistats package is not meant to perform spatial preprocessing, but only statistical analyses on the voxel times series, regardless of the coordinate system.
```
kchawla-pi: Once this is done,

kchawla-pi:
At many places there is improper use of the English articles. Examples:

"decomposed into a voxels (the equivalent" should be "decomposed into voxels (the equivalent"

"forming an a haemodynamic response which " should be "forming a haemodynamic response which "

doc/introduction.rst
kchawla-pi: Before fMRI images can be used to do meaningful comparisons, they must be processed to ensure that the voxels that are being compared represent the same brain regions, irrespective of the variability in size and shape of the brain and its microarchitecture across different subjects in the experiment.
The process is called spatial registration or spatial normalization. During this procedure, the voxels of all the brain images are 'registered' to correspond to the same region of the brain. Usually, the images (their voxels) are registered to a standard 'template' brain image (its voxels) . One often used standard template is the MNI152 template from the Montreal Neurological Institute.
One this is done, the coordinates of a voxel are in the same space as the template and can be used to estimate its brain location using brain atlases based on tht same template.
As already mentioned, the nistats package is not meant to perform spatial preprocessing, but only statistical analyses on the voxel times series, regardless of the coordinate system.

Issue #247, opened 2018-10-05
"singular matrix" occurs too often in the examples/tests


bthirion:
Too many warnings "Attention: Design matrix is singular. Aberrant estimates are expected."

PR #265, opened 2018-10-08
[MRG] Fixed erroneous tests

bthirion:
Addresses #247
Realized that some tests were mistakenly producing singular design matrices, hence were doing fitting noise ;-(


bthirion:
Mergin now.

PR #287, opened 2018-11-26
FIX: numpy.linspace parameter num must be integer

effigies:
`np.ceil` is a conservative interpretation of a float for number of elements to create with `linspace`, ensuring an oversampling rate of at least that specified. If approximate is all that's needed, then `floor` might save a little computation, and `rint` would be the best approximation.

Also proposed in bids-standard/pybids#302.


effigies:
Given that `ceil` breaks tests, I'm assuming we've been implicitly relying on rounding. Switching to `rint`.

Issue #284, opened 2018-11-19
Mutable default paramter in first_level_model_from_bids


kchawla-pi:
In `nistats.first_level_model.first_level_model_from_bids`, here is the function signature:

`first_level_models_from_bids( task_label, space_label, img_filters=[],...`

Why is a mutable data structure (`[]`) the default arg for `img_filter`? This is generally the cause of hard to diagnose bugs in Python. Any particular reason it has been written this way?

bthirion:
No good reason I believe. We should use `img_filters=None`

PR #289, opened 2018-11-29
[MRG] Fixed: param img_filters no longer has a mutable arg ([]) as default

kchawla-pi:
For issue #284 
Mutable default args are a source of hard to diagnose bugs. This fixes it.


bthirion:
LGTM, thx !

Issue #235, opened 2018-09-26
Nistats has no continuous testing support for Windows platform


kchawla-pi:
We should add Appveyor for Nistats, the way we have for Nilearn.
Nistats is planned to be merged with Nilearn & Nistats test must work on AppveyorCI before the merge.
If not, Nilearn AppveyorCI may start failing post merge & might be more complex to fix than in a small separate library such as Nistats.

bthirion:
Good point -- though not an emergency. B

kchawla-pi:
Most of the green sections do not. need to be reviewed, they are full of unmodified code which has been slightly restructured but is showing up as new additions in the Github diff.

I will describe what I did & why, and clarify the parts that actually need to be reviewed.

PR #290, opened 2018-11-29
[MRG] Added support for AppveyorCI for windows bulds

kchawla-pi:
for issue #235 
Based on Nilearn's appveyor.yml
The work has been migrated to PR #306


kchawla-pi:
@bthirion lots of failures in Appveyor. Some of them Windows specific (PermissionError: Files in windows can't be accessed by two processes simultaneously, unlike in Unix)
We should look at these before we can Nistats+Nilearn.

bthirion:
Yes. I think that you're in a good position to address these issues ;-)
Thx !

kchawla-pi:
@GaelVaroquaux @bthirion The tempfile module's constructs seem to be raising the PermissionErrors in Windows. Switching them with manual try-except-finally to manage temp files & dirs seem to work.

Nilearn doesn't seem to be using Python's tempfile module at all. This could be why.
The most straightforward solution seems to do what I mentioned above. You guys have better ideas?

bthirion:
can one use the `write_tmp_imgs` of Nilearn then ?

nistats/tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -48,16 +51,18 @@ def _run_test_for_invalid_separator(filepath, delimiter_name):
...
-                                            delimiter_name=delimiter_name)
+        tmp_delimited_filename = 'tmp_{}_separated_values_file'.format(delimiter_name)
+        try:
+            with open(tmp_delimited_filename, 'w') as temp_csv_obj:
+                _create_test_file(temp_csv=temp_csv_obj,
+                                  test_data=data_for_temp_datafile,
+                                  delimiter=delimiter_char)
+                _run_test_for_invalid_separator(filepath=temp_csv_obj.name,
+                                                delimiter_name=delimiter_name)
+        except:
```
aabadie: all these `except: raise` are not needed.

nistats/tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -93,23 +98,33 @@ def test_binary_opening_an_image():
...
-        with assert_raises(ValueError):
-            result = _verify_events_file_uses_tab_separators(
-                    events_files=temp_bin_obj.name)
-
+    tmp_bin_filename = 'tmp_binary_file.bin'
+    try:
+        with open(tmp_bin_filename, 'wb') as temp_bin_obj:
+            temp_bin_obj.write(temp_data_bytearray_from_ints)
+            with assert_raises(ValueError):
+                result = _verify_events_file_uses_tab_separators(
```
aabadie: `result` is not used after.

nistats/tests/test_first_level_model.py
```diff
@@ -7,6 +7,12 @@
 
 import os
 import shutil
+
+try:
+    from tempfile import TemporaryDirectory
+except ImportError:
+    from nistats.utils import TemporaryDirectory
```
aabadie: In which case do you need your version of `TemporaryDirectory` ? A comment would be welcome here.

nistats/tests/test_first_level_model.py
```diff
@@ -76,7 +80,7 @@ def test_high_level_glm_one_session():
 
 def test_high_level_glm_with_data():
     # New API
-    with InTemporaryDirectory():
+    with TemporaryDirectory():
```
aabadie: What is the problem with `InTemporaryDirectory` from nibabel ?

nistats/tests/test_first_level_model.py
```diff
@@ -221,6 +225,7 @@ def test_fmri_inputs():
             # confounds rows do not match n_scans
             assert_raises(
                 ValueError, FirstLevelModel(mask=None).fit, fi, d, conf)
+            import time
```
aabadie: leftover

nistats/tests/test_verify_events_file_uses_tab_separators.py
```diff
@@ -93,23 +98,33 @@ def test_binary_opening_an_image():
     img_data = bytearray(
             b'GIF87a\x01\x00\x01\x00\xe7*\x00\x00\x00\x00\x01\x01\x01\x02\x02'
             b'\x07\x08\x08\x08\t\t\t\n\n\n\x0b\x0b\x0b\x0c\x0c\x0c\r;')
-    with NamedTemporaryFile(mode='wb', suffix='.gif',
-                            dir=os.getcwd()) as temp_img_obj:
-        temp_img_obj.write(img_data)
-        with assert_raises(ValueError):
-            _verify_events_file_uses_tab_separators(
```
aabadie: So looking more at the code, I think the problem is more related to how `_verify_events_file_uses_tab_separators` works: internally it reopens the same file which is forbidden on Windows.

In the test, the `NamedTemporaryFile` returns a file object (already opened) that you try to reopen in `_verify_events_file_uses_tab_separators` and here windows is unhappy. Your proposed solution suffers the same problem but it's worse because you hide it using a try/except.

I would suggest to refactor `_verify_events_file_uses_tab_separators` to accept file object parameters and when used not reopen it. I think this will avoid lots of changes proposed by this PR.

nistats/tests/test_verify_events_file_uses_tab_separators.py
kchawla-pi: I replaced the `with NamedTemporaryFile() as ...` with try-except-finally to manually manage the file close and remove. The latter approach was not generating the errors the former was. 
You are suggesting it is the function's implementation that is the source of the problem, which is good news. I will work with you to address the implementation and then restore the former.

nistats/tests/test_verify_events_file_uses_tab_separators.py
kchawla-pi: I have looked at the code and here are some thoughts:

> I think this will avoid lots of changes proposed by this PR.

There were ~13 tests failing on Windows and this accounts for only 3-4 of them. The other changes I have made are not because of this function.

> I would suggest to refactor _verify_events_file_uses_tab_separators to accept file object parameters ...

It maybe that the function could have used better design principles but this was not caught when it was being written.
Refactoring it now would also mean some code modification everywhere the function is being used, and that would & should be a separate PR, not this one. I recognize that this code fragility is symptomatic of bad code design, and I will address it immediately, just not in this PR.

It seems to me backwards to rewrite the testee function to make the tests work with it, instead of modifying the tests to work with the function being tested. The only immediate benefit is that we will continue using the `with` statement instead of manually doing a `try-except-finally`, which is what `with` essentially does.

I propose that we keep these changes here so we have testing on Windows platform for subsequent PRs which will be forthcoming soon & open a new PR to redesign this function & associated tests.

kchawla-pi:
@aabadie I have addressed all the comments. Not the refactor of the underlying function, that will be a separate PR. 
This PR only enables Windows testing for Nistats which I want enabled ASAP before any more PRs are submitted or merged, including the new ones I have to start on now.

kchawla-pi:
I examined the code further and gained some more insight about the problems in using nibabel's `InTemporaryDirectory` and a potential solution.
I will see if I can incorporate that in this PR.

bthirion:
Say I am OK to merge the code, but I'd liek to have @aabadie's agreement.
We should also properly flag the changes to be performed in an issue.
Thx  to all of you !

kchawla-pi:
I rewrote pretty much the entire PR and I think this is the best approach so far.

I wanted to create a decorator for the InTemporaryDirectory but it keeps giving me errors. ANy insight into that would be great.

kchawla-pi:
Alright! AFAIC, this PR is ready to deploy. 
Succeeded in turning nibabel's `InTemporaryDirectory` class into an easy to use decorator.
 - I have written it so it does not muddy the stack traces. Users will be able to debug without having to deal with the decorator itself-the stack trace will point to the actual function only, not the decorator's code.
 - It keeps the test code clean.
 - It is easy to read, use & understand.
 - It can now be used in future tests to make sure errors don't crop up on Windows when using temporary directories.

kchawla-pi:
@aabadie Hope you had a good vacation! Time to review this! :-)

nistats/first_level_model.py
```diff
@@ -397,7 +397,7 @@ def fit(self, run_imgs, events=None, confounds=None,
                 self.masker_.fit(run_imgs[0])
             else:
                 self.masker_ = self.mask
-
+
```
aabadie: Trailing whitespaces added ?

nistats/tests/test_first_level_model.py
```diff
@@ -7,20 +7,32 @@
 
 import os
 import shutil
+
```
aabadie: all those changes in the imports look unrelated ? Or is it just because you want to import `with_temp_dir` ?

nistats/tests/test_first_level_model.py
```diff
@@ -144,8 +131,8 @@ def test_high_level_glm_null_contrasts():
     z2 = single_session_model.compute_contrast(np.eye(rk)[:1],
                                                output_type='stat')
     np.testing.assert_almost_equal(z1.get_data(), z2.get_data())
-
-
+
```
aabadie: whitespaces

kchawla-pi:
> The within_temp_dir decorator is nice but could it be added in a follow-up PR

I don't think so, it is part of getting the tests to work on Windows. I do understand the number of changes look huge, however that is misleading due to the way Github shows diffs.

Most of the green sections do not. need to be reviewed, they are full of existing code which has been slightly restructred but is showing up as new additions in the Github diff.

I will describe what I did & why, and clarify the parts that actually need to be reviewed.

This is how the tests were written in `master`:
```
def test_function():
    with InTemporaryDirectory():
        <code to open some mock files, run tests etc>
```
The problem was that when exiting the `with` context in these tests, the opened mock files would still be open, so the attempt of `InTemporaryDirectory` to delete the temp dir & its contents would fail in Windows, since Windows does not permit moving or deleting files with open handles.

One solution was to either track down all the files that stayed open and manually close them before exiting the context, which some tests did.

I found that to be onerous and cluttered, so I restructured the tests thus:

```
def _code_for_testing_function():
        <code to open some mock files, run tests etc>

def test_function():  # Necessary to put it in a wrapping function for testing frameworks to find it.
    with InTemporaryDirectory():
        _code_for_testing_function()
```

This solved the problem nicely. When `_code_for_testing_function()` would complete execution, the function would go out of scope and all open file handles would be closed.

This is what I did first (after I rewrote the entire PR). But it looked super cluttered, ugly, hard to read.

Which is when I created the decorator to replace `with InTemporaryDirectory():`
```
@within_tmp_dir
def test_function():
        <code to open some mock files, run tests etc>
```
This solved both the problems.

So most of the code changes in the file are just this. The test code itself has not been altered. Merely restructured.

Does that help? @aabadie

kchawla-pi:
The work has been migrated to PR #306 .

PR #296, opened 2018-12-06
Added a DeprecationWarning for Python 3.4

kchawla-pi:
Same as issue https://github.com/nilearn/nilearn/issues/1877
Also replaced use of six with sys.version_info for Python version determination (stdlib use over ext lib).

### DIFF
--- a/nistats/__init__.py
+++ b/nistats/__init__.py
@@ -32,7 +32,7 @@ import warnings
 from .version import _check_module_dependencies, __version__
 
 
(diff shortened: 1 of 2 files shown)


bthirion:
LGTM, thx.

kchawla-pi:
@bthirion  Review & Finalize but don't do any merges yet. I am exploring some merges on Nistats and if we should revert them and I don't want to complicate it if it comes to that.

PR #306, opened 2019-01-18
[MRG+1] Fix Nistats test to run on Windows (using Appveyor CI)

kchawla-pi:
for issue #235
Based on Nilearn's appveyor.yml
In lieu of PR #290 

This PR was extensively reworked and the thread can be started at https://github.com/nistats/nistats/pull/306#issuecomment-458947516


kchawla-pi:
Most of the green sections do not. need to be reviewed, they are full of existing code which has been slightly restructured but is showing up as new additions in the Github diff.

I will describe what I did & why, and clarify the parts that actually need to be reviewed.

kchawla-pi:
I am resorting & reformatting the imports to be PEP8 compliant.
I am doing these small fixes where I notice them because none of us are likely to open PRs specifically addressing these, or notice these except when working on another PR. Not making these small improvements will mean the code base will look more and more non-compliant and harder to read.
I think the slightly increased cognitive load is worth it.

kchawla-pi:
This is how the tests were written in `master`:
```
def test_function():
    with InTemporaryDirectory():
        <code to open some mock files, run tests etc>
```
The problem was that when exiting the `with` context in these tests, the opened mock files would still be open, so the attempt of `InTemporaryDirectory` to delete the temp dir & its contents would fail in Windows, since Windows does not permit moving or deleting files with open handles.

One solution was to either track down all the files that stayed open and manually close them before exiting the context, which some tests did.

I found that to be onerous and cluttered, so I restructured the tests thus:

```
def _code_for_testing_function():
        <code to open some mock files, run tests etc>

def test_function():  # Necessary to put it in a wrapping function for testing frameworks to find it.
    with InTemporaryDirectory():
        _code_for_testing_function()
```

This solved the problem nicely. When `_code_for_testing_function()` would complete execution, the function would go out of scope and all open file handles would be closed.

This is what I did first (after I rewrote the entire PR). But it looked super cluttered, ugly, hard to read.

Which is when I created the decorator to replace `with InTemporaryDirectory():`
```
@within_tmp_dir
def test_function():
        <code to open some mock files, run tests etc>
```
This solved both the problems.

So most of the code changes in the file are just this. The test code itself has not been altered. Merely restructured.

Pretty much the entire change set in `nistats/tests/test_first_level_model.py` and `nistats/tests/test_second_level_model.py` are of this nature.

kchawla-pi:
nibabel's `InTemporaryDirectory` class was turned into an easy to use decorator.

I tried many approaches and this was the one which provided the cleanest, easiest to read & use, effective solution. It can now be used when writing new tests to make sure errors don't crop up on Windows when using temporary directories.

I have written it so it does not muddy the stack traces. Users will be able to debug without having to deal with the decorator itself-the stack trace will point to the actual function only, not the decorator's code.

kchawla-pi:
@aabadie 
@bthirion

nistats/tests/test_datasets.py
```diff
@@ -171,26 +167,26 @@ def _expected_output_data_from_test_file():
...
         return file_data
     
+    @within_temp_dir
     def run_test():
         data_for_tests = _input_data_for_test_file()
         expected_data_from_test_file = _expected_output_data_from_test_file()
-        with NamedTemporaryFile(mode='w',
-                                dir=os.getcwd(),
-                                suffix='.csv') as temp_csv_obj:
-            data_for_tests.to_csv(temp_csv_obj.name,
```
aabadie: Using `.name` attribute is not needed since `to_csv` function also works with file objects (e.g temp_csv_obj). This will avoid a reopening of the same file and probably fix the related windows issue (multiple file access are forbidden).

nistats/tests/test_datasets.py
```diff
@@ -171,26 +167,26 @@ def _expected_output_data_from_test_file():
...
-                                  index=False,
-                                  header=False,
-                                  sep=' ',
-                                  )
-            datasets._make_events_file_localizer_first_level(
-                    temp_csv_obj.name
-                    )
-            data_from_test_file_post_mod = pd.read_csv(temp_csv_obj.name,
```
aabadie: Same here

kchawla-pi:
Following discussions with Alexandre Abadie, the plan is to make more granular, piecemeal, minimum changes for this PR.

nistats/tests/test_datasets.py
```diff
@@ -174,18 +175,15 @@ def _expected_output_data_from_test_file():
     def run_test():
         data_for_tests = _input_data_for_test_file()
         expected_data_from_test_file = _expected_output_data_from_test_file()
-        with NamedTemporaryFile(mode='w',
-                                dir=os.getcwd(),
-                                suffix='.csv') as temp_csv_obj:
-            data_for_tests.to_csv(temp_csv_obj.name,
+        temp_tsv_file = 'temp_file.tsv'
+        with InTemporaryDirectory():
```
aabadie: You are not following my suggestion, what's wrong with it ?

kchawla-pi: I had changed temp_csv_obj.name with temp_csv_obj as you suggested and it was working fine. However, an error unrelated to that happened that it was not reading the csv properly into a pandas dataframe. Skipping NamedTemporaryFile easily solved this with a one line change..

kchawla-pi:
@aabadie There might be something wrong with the test files that are being created. Don't let that stop you from figuring this out, while I continue looking into this.

kchawla-pi:
The f.seek(0) is not helping and it would seem the test file is not being written to correctly.
It is occurring to me that we should use nibabel.InTemporaryDirectory, since that approach mimic the actual use case correctly.
Opening a file handle and writing, then resetting seek to 0, reading, then resetting seek to 0,  and so on is not how the function being tested is actually supposed to work.

kchawla-pi:
Pandas seems to be having trouble writing to open file handles.
FOund this also: https://github.com/pandas-dev/pandas/issues/19827

kchawla-pi:
@aabadie I think we should fix the function and the test first, merge that and then continue with this?

kchawla-pi:
Not sure why but when doing all this seek(0) stuff the values are not being written properly.

kchawla-pi:
@aabadie Ready for another review.

bthirion:
This PR looks somewhat empty now. Is it to be closed ?
Best,
B

kchawla-pi:
No,it's about ready now. Awaiting a final green from @aabadie and you and then I will merge it.

kchawla-pi:
It's empty cuz I worked with @aabadie to not change more than was necessary for this PR. I have filed other PRs for the unrelated changes.

nistats/utils.py
```diff
@@ -119,10 +120,21 @@ def _verify_events_file_uses_tab_separators(events_files):
...
-                    'The values in the events file are not separated by tabs; '
-                    'please enforce BIDS conventions',
-                    events_file_)
-    
+                        'The values in the events file '
+                        'are not separated by tabs; '
+                        'please enforce BIDS conventions',
+                        events_file_
+                        )
+            try:
```
aabadie: I don't understand why this block is needed now but not before.

kchawla-pi: I was going through the docs and saw the header method and occurred to me a binary file will have readable tab separated headers, a violation of BIDS protocol, so it serves this function's purpose.

kchawla-pi:
@aabadie Seems its all working now. Check it out.

kchawla-pi:
@bthirion time for your review. I intend to merge this ASAP once the CI is green.

PR #313, opened 2019-02-01
[WIP] Refactor & improve functions & tests related to BIDS compatibility of events files

kchawla-pi:
Unnests some nested helper functions in tests.
Modifies _make_events_file_localizer_first_level() to work correctly with file handles.


kchawla-pi:
@aabadie This is the one where I will fix & redo the other functions. Please work along me with review etc.

kchawla-pi:
Generally yes but this is an internal function for an internal function/test, I want it to be unambiguous about which internal function is helping which test.

kchawla-pi:
There are 3 tests that are now failing on Windows (not raising ValueError in windows, as they do in Linux). I may have to change the tests or the function or both. What are your thoughts on that?

kchawla-pi:
Ignore the appveyor error, it was not even supposed to run in this PR.

Issue #318, opened 2019-02-20
incorrect documentation for period_cut


poldrack:
The documentation for nistats.first_level_model.FirstLevelModel say that period_cut is for a low-pass filter:

```
period_cut : float, optional

    This parameter specifies the cut period of the low-pass filter in seconds for the design matrices.

``` 

when i fact it's for a high-pass filter.  it should also make clear that this is only for the cosine option.

bthirion:
Indeed, thx.

PR #319, opened 2019-02-26
[MRG] fixed 3 dosctrings

bthirion:
Address #318.

### DIFF
--- a/nistats/design_matrix.py
+++ b/nistats/design_matrix.py
@@ -86,10 +86,10 @@ def _cosine_drift(period_cut, frame_times):
     Parameters
     ----------
     period_cut : float
-        Cut period of the high-pass filter (in sec)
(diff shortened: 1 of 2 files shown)


bthirion:
Please one review, so that we can merge it.

nistats/design_matrix.py
```diff
@@ -86,7 +86,7 @@ def _cosine_drift(period_cut, frame_times):
     Parameters
     ----------
     period_cut : float
-        Cut period of the low-pass filter (in sec)
+        Cut period of the high-pass filter (in sec)
```
kchawla-pi: We should make the `in sec ` consistent one way or another across these docstrings.

kchawla-pi:
From what was mentioned in #318 , The changes do not mention that the high-pass filter option only works with cosines.

Issue #293, opened 2018-11-30
Discrepancy in terms used between Nistats & Nilearn.


kchawla-pi:
NiftiMasker defined in FirstLevelModel might be different between NiStats & NiLearn.
Terms to check:
t_r
hi_frequency_cutoff
lo_frequency_cutoff

@bthirion Anna-Louisa @alpinho might be interested in working on this as well. Any helpful info to flesh out this issue should be added here.

bthirion:
@alpinho do you consider  working on it ?

alpinho:
Yes! I will work on it next week.

kchawla-pi:
@bthirion  It occurred to me that it is not straightforward to maintain backward & forward compatibility with this change, based on the changes @alpinho has made so far in her PR #322 . 
Making changes to preserve backward compatibility will not be clean. And once the deprecation period is past, changing it to look clean for future will not be clean.
I may be able to engineer around this using decorators, but I don't think the benefit is worth the cost.

I am thinking that we either forego backward compatibility, or that we make this change in `nilearn.stats` instead and leave this unchanged here

bthirion:
Yes, for Nistats we do not ensure backward compatibility atm.
Please don't add decorators ;-)
B

kchawla-pi:
Cool I will ask Ann-Luisa to just make the changes, that should simplify everything.

PR #322, opened 2019-03-11
ENH: added attributes low and high pass to FLM for parity with Nilearn NiftiMasker

alpinho:
Solving issue #293


nistats/first_level_model.py
```diff
@@ -261,6 +261,14 @@ class FirstLevelModel(BaseEstimator, TransformerMixin, CacheMixin):
         This id will be used to identify a `FirstLevelModel` when passed to
         a `SecondLevelModel` object.
 
+    low_pass: None or float, optional
+        This parameter is passed to signal.clean. Please see the related
+        documentation for details. [1]
```
bthirion: Sorry, i don' like such reference is dostrings, as I usually look at them in interactive mode.

nistats/first_level_model.py
```diff
@@ -315,6 +328,8 @@ def __init__(self, t_r=None, slice_time_ref=0., hrf_model='glover',
         self.labels_ = None
         self.results_ = None
         self.subject_label = subject_label
+        self.low_pass = low_pass
+        self.high_pass = high_pass
```
bthirion: But then, you need to test that the new behavior is enforced. and you should remove the 'period_cut' argument as it is redundant with high_pass.

kchawla-pi: We could not write tests, her computer had no space to install nosetests and it's dependencies.

nistats/first_level_model.py
kchawla-pi: Is that true for all users? I use IDEs and appreciate if I can click on a link in the docstring for more details instead of typing it in a browser.

I am sure there are many who will do it your way and many who will like the hyperlink.

Also, does having the hyperlink interfere with the help display in interactive mode somehow? If not, then  having this link is compatible with both preferences.

alpinho:
In the design_matrix.py file, the period_cut is converted in frequency on the line 103 (inside the function _cosine_drift):

https://github.com/nistats/nistats/blob/e6621dac6d5dc6507d553b8c1c0ae471b8f4676c/nistats/design_matrix.py#L103

We can state hfcut=high_pass and set the default value of high_pass=0.01Hz in the class FirstLevelModel inside first_level_model.py

We will also deprecated the period_cut parameter. 
@bthirion

kchawla-pi:
Here're the factors for our consideration and what we plan to do about them:

1. Several private functions have a default arg for `period_cut`. We will remove the default args, since they make maintenance harder and functionality more opaque. Private functions should not have any default options at all.

2. We set the default arg for `period_cut` in `FirstLevelModel` & `make_first_level_model_design_matrix` to `None`, and add `high_pass=0.01`, `low_pass=None`. If the user specifies a `period_cut` value, we  convert the `period_cut` value to Hz and pass it to `high_pass`.

3. If a user manually sets values for both `period_cut` & `high_pass` (and this will be new code, not existing code), `low_pass`, we raise an error telling them to choose `high_pass`& `low_pass`.

4. We issue a DeprecationWarning, advising against the use of `period_cut`.

bthirion:
For 3. I'd not raise an error. I would spit a wrning, and only take into account the high_pass value.

LGTM otherwise.

alpinho:
We have run the tests and we get an error in the test_dmtx.py, line 183. len(names) = 4, but it shall be 8. We believe the 4 drift regressors are not properly defined anymore because we have set low_pass to None. Shall we also have a default value for low_pass? What about 0.08Hz?

bthirion:
.08hz is probably the closest match to what people do in practice.
+1 for it.
B

kchawla-pi:
It just occurred to me that in Python 2 `/` does an internet division not a
floating point division. So using `.`  to convert the existing value of
period cut may also be necessary.
Let's try that as well before setting a default `low_pass` value.

alpinho:
Yes, it seems that was the problem. I changed it in both first_level_model.py and design_matrix.py and the error disappeared. I got a new one, though, inside the function ```test_cosine_drift```:

https://github.com/nistats/nistats/blob/e6621dac6d5dc6507d553b8c1c0ae471b8f4676c/nistats/tests/test_dmtx.py#L100

Here the traceback:

```
======================================================================
FAIL: nistats.tests.test_dmtx.test_cosine_drift
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/analu/nistasts_py27_dev/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/volatile/mygit_repos/nistats/nistats/tests/test_dmtx.py", line 101, in test_cosine_drift
    assert_almost_equal(spm_drifts[:, 1:], nistats_drifts[:, : -2])
  File "/home/analu/nistasts_py27_dev/local/lib/python2.7/site-packages/numpy/testing/_private/utils.py", line 572, in assert_almost_equal
    return assert_array_almost_equal(actual, desired, decimal, err_msg)
  File "/home/analu/nistasts_py27_dev/local/lib/python2.7/site-packages/numpy/testing/_private/utils.py", line 1007, in assert_array_almost_equal
    precision=decimal)
  File "/home/analu/nistasts_py27_dev/local/lib/python2.7/site-packages/numpy/testing/_private/utils.py", line 752, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 7 decimals

(shapes (20, 3), (20, 399) mismatch)
 x: array([[ 0.3152529,  0.3123345,  0.3074904],
       [ 0.3074904,  0.281761 ,  0.2404615],
       [ 0.2921564,  0.2236068,  0.1210151],...
 y: array([[ 0.3152529,  0.3123345,  0.3074904, ...,  0.3074904,  0.3123345,
         0.3152529],
       [ 0.3074904,  0.281761 ,  0.2404615, ...,  0.2404615,  0.281761 ,...
```

I also got some UserWarnings that might be related:

https://github.com/nistats/nistats/blob/e6621dac6d5dc6507d553b8c1c0ae471b8f4676c/nistats/design_matrix.py#L501

https://github.com/nistats/nistats/blob/e6621dac6d5dc6507d553b8c1c0ae471b8f4676c/nistats/utils.py#L276

nistats/design_matrix.py
```diff
@@ -79,14 +79,13 @@ def _poly_drift(order, frame_times):
     return pol
 
 
-def _cosine_drift(period_cut, frame_times):
-    """Create a cosine drift matrix with periods greater or equal to
-    period_cut.
+def _cosine_drift(high_pass, frame_times):
+    """Create a cosine drift matrix with frequencies equal or greater high_pass.
```
chrplr: line 83 : high_pass -> than "high_pass"

nistats/tests/test_dmtx.py
```diff
@@ -43,6 +43,7 @@ def design_matrix_light(
                                           drift_model, period_cut, drift_order, fir_delays,
                                           add_regs, add_reg_names, min_onset)
     _, matrix, names = check_design_matrix(dmtx)
+    print matrix, names
```
chrplr: line 46: print will crash in python3, no?

bthirion: Also, we should not print in tests.

bthirion:
@alpinho do you plan to finish the PR ?

alpinho:
> @alpinho do you plan to finish the PR ?

Yes! I am.

alpinho:
Closed in favour of PR #351

PR #326, opened 2019-04-02
[WIP] Unvendor sphinx-gallery

kchawla-pi:
Issue #295

PR #327, opened 2019-04-02
[WIP] Speed up python env setup in CircleCI & Appveyor

kchawla-pi:
Small optimizations to speed up CircleCI:
 - Default conda is 3.7, abandon 3.6 so new python doesn't have to be downloaded and installed.
 - Using conda-forge to install non-default channel packages upgrades all conda installed packages. Switch to using pip for these instead.

Small optimizations to speed up Appveyor:
 - Use pip for installation of packages instead of conda-forge channel.
- Remove the wheel build step.


bthirion:
Same questions as for #326

kchawla-pi:
No, the experiments did not you'll any benefit. I'll probably closer this
soon

PR #361, opened 2019-06-18
[NOMRG] add xml reports

jeromedockes:
this branch is not meant to be merged but just try a few things and help
discussion around #352.

here reports are generated as XML, which makes it easy to store a lot of
information (this is just a toy prototype) that can be easily validated (e.g. in
tests) for example with schemas, and reused.

xslt is used to provide human-readable views of the reports.

in other projects I have found that the pattern (info stored in xml + views
generated with XSLT + css to make them pretty) can be very efficient.

things that can be reused in #352:

- even if not using xml, lxml (which can work with html) can be useful to
  construct the reports

- store matplotlib figures as svg rather than binary formats such as png. note:
  to generate valid xhtml make sure ids of elements generated by matplotlib are
  actually unique in the document by changing them in the svg.

- remove deprecated "border" attribute written by pandas in html tables

### DIFF
--- a/nistats/reporting/report_template.xml
+++ b/nistats/reporting/report_template.xml
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<nistats:report xmlns:nistats="https://nistats.github.io">
+    <nistats:title></nistats:title>
+    <nistats:model_info></nistats:model_info>
+    <nistats:contrast_list></nistats:contrast_list>
+    <nistats:design_matrices></nistats:design_matrices>
+</nistats:report>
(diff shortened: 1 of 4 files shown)


kchawla-pi:
This is pretty neat Jerome, and a very nice work. Thanks! I'll run the example and see the report.

effigies:
Might be worth having a look at [NIDM-Results](https://nidm.nidash.org/specs/nidm-results_130.html) if you're going down this road.

jeromedockes:
> Might be worth having a look at [NIDM-Results](https://nidm.nidash.org/specs/nidm-results_130.html) if you're going down this road.

indeed if we decide to create xml reports we will be using `nidm` elements rather than those used in this toy example. the reports being reusable by tools that understand this language would be a very good thing

Issue #378, opened 2019-08-25
Support missing regressors in FirstlevelGLM 


bthirion:
Currently, the occurrence of different sets of trial types across sessions in a FirstLevelGLM raises an error, because the model expects the conditions to be identical in all sessions.
This is an issue with randomized designs, where one of the conditions is absent.

Alternatively ---or as a first step ?--- Nistats should provide an example where this is handled with a separate fit of the different sessions and then an explicit fixed effects analysis.

Issue #350, opened 2019-05-14
first level uses only the first design matrix


satra:
> What version of Nistats are you using? 
i was looking at master

> What were you trying to do? 
understand how nistats is modeling run level data

> What did you expect will happen?
that the designs of each run are incorporated

> What actually happened?
only first run design is being used

see: https://github.com/nistats/nistats/blob/master/nistats/first_level_model.py#L552

if the designs are different across runs (see some reasons below), this would not generate appropriate contrasts.

reasons for designs being different across runs:
- number of outliers being included per run
- missing conditions (in a hit/miss experiment, it's possible there are no hits or misses)
- number of noise components (e.g., compcor) being included to explain variance could vary per run

cc: @effigies

bthirion:
Thx for reporting.
If I understand correctly, you want to run a fixed effects analysis on a multi-session dataset, with some missing conditions in some of the sessions ?
I have no good solution for that in the library: what you can do easily is to run independent models on the different sesssions, but then there exists no function that runs the fixed effects analysis on the analysis.
You probably need something like 
https://github.com/hbp-brain-charting/public_analysis_code/blob/master/ibc_public/utils_pipeline.py
l.557-598.
?

satra:
@bthirion - thank you for that pointer. i may use that for other things.

but yes, the main thing is that designs can vary by run within a task of a single subject, and contrasts may need to be combined in specific across runs. we could potentially create a nan contrast when it cannot be computed and then in the fixed effects code use nansum and nanmax

AnnaD15:
I'n not sure but I think my problem is relalated to this one. I have 3 session:
- 1 session: viewing faces and rest
- 2 session: viewing pictures and rest
- 3 session: viewing words and rest
I can compare Faces minus Faces_rest with the code below. However this does not work for pictures and words, probably because faces are my first session and thus the design matrix used. To get the contrast for pictures and words I could add the run number in the following line like this: model.fit(imgs[1] , events[1] , confounds[1] ) by adding the [1] for the picture contrast and [2] for the words. 
This way I got the contrast in each session for each condition, however I'm also interested in comparing faces, pictures and words. When I try computing those 
"zmap = model.compute_contrast('Faces-IAPS', 't',output_type=my_output)"
I get the following error message: "PatsyError: unrecognized token in constraint Faces-IAPS", probably because nistats only makes use of the first design matrix.
What is the correct way to compare those three sessions in Nistats?

Thanks a lot for any recommendations!

```
[...]

PR #386, opened 2019-09-06
[MRG] introduced fixed effects model

bthirion:
Addresses #378 #350 #281

### DIFF
--- a/examples/02_first_level_models/plot_fixed_effects.py
+++ b/examples/02_first_level_models/plot_fixed_effects.py
@@ -13,7 +13,7 @@ JB. Functional segregation of cortical language areas by sentence
 repetition. Hum Brain Mapp. 2006: 27:360--371.
 http://www.pubmedcentral.nih.gov/articlerender.fcgi?artid=2653076#R11
 
-Please see :doc:`Simple example of two-session fMRI model fitting
+Please see `Simple example of two-session fMRI model fitting
 <https://nistats.github.io/auto_examples/02_first_level_models/plot_fiac_analysis.html>`
 example for details.  The main difference is that
 the fixed-effects model is run explicitly here,
--- a/examples/02_first_level_models/plot_fixed_effects.py
+++ b/examples/02_first_level_models/plot_fixed_effects.py
@@ -1,8 +1,9 @@
-"""
-Example of explicit fixed effects fMRI model fitting
+"""Example of explicit fixed effects fMRI model fitting
 ====================================================
 
-This example illustrates how to 
+This example illustrates how to run a fixed effects model based on
+pre-computed statistics. This is helpful when the initial models
+have to be fit separately.
 
 For details on the data, please see:
 
@@ -12,15 +13,15 @@ JB. Functional segregation of cortical language areas by sentence
 repetition. Hum Brain Mapp. 2006: 27:360--371.
 http://www.pubmedcentral.nih.gov/articlerender.fcgi?artid=2653076#R11
 
-Please see `plot_fiac_analysis.py` example for details.  The main
+Please see `../plot_fiac_analysis.py` example for details.  The main
 difference is that the fixed-effects model is run explicitly here,
 after GLM fitting on two sessions.
 
 """
 
 ###############################################################################
-# Create a write directory to work
-# it will be a 'results' subdirectory of the current directory.
+# Create a write sub-directory named 'results'
+# 
 from os import mkdir, path, getcwd
 write_dir = path.join(getcwd(), 'results')
 if not path.exists(write_dir):
@@ -30,7 +31,7 @@ if not path.exists(write_dir):
 # Prepare data and analysis parameters
 # --------------------------------------
 # 
-# Note that there are two sessions
+# Inspecting 'data', we note that there are two sessions
 
 from nistats import datasets
 data = datasets.fetch_fiac_first_level()
@@ -78,26 +79,32 @@ from nilearn import plotting
 cut_coords = [-129, -126, 49]
 
 fmri_glm = fmri_glm.fit(fmri_img[0], design_matrices=design_matrices[0])
-dic1 = fmri_glm.compute_contrast(contrast_val, output_type='all')
+summary_statistics_session1 = fmri_glm.compute_contrast(
+    contrast_val, output_type='all')
 plotting.plot_stat_map(
-    dic1['z_score'], bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
+    summary_statistics_session1['z_score'],
+    bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
     title='%s, first session' % contrast_id)
 
 #########################################################################
 # Statistics for the second session
 
 fmri_glm = fmri_glm.fit(fmri_img[1], design_matrices=design_matrices[1])
-dic2 = fmri_glm.compute_contrast(contrast_val, output_type='all')
+summary_statistics_session2 = fmri_glm.compute_contrast(
+    contrast_val, output_type='all')
 plotting.plot_stat_map(
-    dic2['z_score'], bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
+    summary_statistics_session2['z_score'],
+    bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
     title='%s, second session' % contrast_id)
 
 #########################################################################
 # Fixed effects statistics
 from nistats.contrasts import fixed_effects_img
 
-contrast_imgs = [dic1['effect_size'], dic2['effect_size']]
-variance_imgs = [dic1['effect_variance'], dic2['effect_variance']]
+contrast_imgs = [summary_statistics_session1['effect_size'],
+                 summary_statistics_session2['effect_size']]
+variance_imgs = [summary_statistics_session1['effect_variance'],
+                 summary_statistics_session2['effect_variance']]
 
 ffx_contrast, ffx_variance, ffx_stat = fixed_effects_img(
     contrast_imgs, variance_imgs, data['mask'])
--- a/examples/02_first_level_models/plot_fixed_effects.py
+++ b/examples/02_first_level_models/plot_fixed_effects.py
@@ -1,5 +1,6 @@
-"""Simple example of fixed effects fMRI model fitting
-================================================
+"""
+Example of explicit fixed effects fMRI model fitting
+====================================================
 
 This example illustrates how to 
 
--- a/examples/02_first_level_models/plot_fixed_effects.py
+++ b/examples/02_first_level_models/plot_fixed_effects.py
@@ -13,8 +13,10 @@ JB. Functional segregation of cortical language areas by sentence
 repetition. Hum Brain Mapp. 2006: 27:360--371.
 http://www.pubmedcentral.nih.gov/articlerender.fcgi?artid=2653076#R11
 
-Please see `plot_fiac_analysis.py` example for details.  The main
-difference is that the fixed-effects model is run explicitly here,
+Please see :ref:`Simple example of two-session fMRI model fitting
+<https://nistats.github.io/auto_examples/02_first_level_models/plot_fiac_analysis.html>`
+example for details.  The main difference is that
+the fixed-effects model is run explicitly here,
 after GLM fitting on two sessions.
 
 """
--- a/examples/02_first_level_models/plot_fixed_effects.py
+++ b/examples/02_first_level_models/plot_fixed_effects.py
@@ -83,21 +83,21 @@ plotting.plot_stat_map(
 
 #########################################################################
 # Fixed effects statistics
-from nistats.contrasts import fixed_effects_img
+from nistats.contrasts import compute_fixed_effects
 
 contrast_imgs = [summary_statistics_session1['effect_size'],
                  summary_statistics_session2['effect_size']]
 variance_imgs = [summary_statistics_session1['effect_variance'],
                  summary_statistics_session2['effect_variance']]
 
-ffx_contrast, ffx_variance, ffx_stat = fixed_effects_img(
+fixed_fx_contrast, fixed_fx_variance, fixed_fx_stat = compute_fixed_effects(
     contrast_imgs, variance_imgs, data['mask'])
 plotting.plot_stat_map(
-    ffx_stat, bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
+    fixed_fx_stat, bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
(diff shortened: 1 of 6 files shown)


bthirion:
@kchawla-pi any hint about windows failure ?

kchawla-pi:
The still in memory objects must be manually deleted using the`del` keyword before exiting the temporary directory context

jeromedockes:
> @kchawla-pi any hint about windows failure ?

the mask NiftiImage holds a file handle to the .nii file on disk in its `file_map`. in windows that file cannot be removed while there is an open file handle

bthirion:
Could I get a review on this PR ? Neurostars is asking for it...

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+    Returns
+    -------
+    ffx_contrast_img: Nifti1Image,
+             the fixed effects contrast computed within the mask
+    ffx_variance_img: Nifti1Image,
+             the fixed effects variance computed within the mask
+    ffx_t_img: Nifti1Image,
+             the fixed effects t-test computed within the mask
+    """
+    from nilearn.input_data import NiftiMasker
```
kchawla-pi: Imports at the top of the file.

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+    if len(contrast_imgs) != len(variance_imgs):
+        raise ValueError(
+            'The number of contrast images (%d) differs from the number of '
+            'variance images (%d). ' % (len(contrast_imgs), len(variance_imgs)))
+
+    # instantiate a masker
+    if isinstance(mask, NiftiMasker):
+        masker = mask
```
kchawla-pi: Are we creating a new object or a new reference to the same object?

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+    else:
+        masker = NiftiMasker(mask_img=mask).fit()
+
+    vars = masker.transform(variance_imgs)
+    cons = masker.transform(contrast_imgs)
+
+    ffx_contrast, ffx_variance, ffx_t = _fixed_effects(
```
kchawla-pi: I would prefer `fixed_fx_`

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
 
     def __div__(self, scalar):
         return self.__rmul__(1 / float(scalar))
+
+
+def fixed_effects_img(contrast_imgs, variance_imgs, mask,
```
kchawla-pi: `make_fixed_effects_img`

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+        cons, vars, precision_weighted)
+
+    ffx_contrast_img = masker.inverse_transform(ffx_contrast)
+    ffx_variance_img = masker.inverse_transform(ffx_variance)
+    ffx_t_img = masker.inverse_transform(ffx_t)
+    return ffx_contrast_img, ffx_variance_img, ffx_t_img
+
+
+def _fixed_effects(contrasts, variances, precision_weighted):
+    """Compute the fixed effets given arrays of effects and variance
```
kchawla-pi: Computes the fixed effects t-statistic, contrast, variance, given arrays of effects and variance.

nistats/tests/test_contrasts.py
```diff
@@ -126,3 +127,19 @@ def test_contrast_values():
     # Note that the values are not strictly equal,
     # this seems to be related to a bug in Mahalanobis
     assert_almost_equal(np.ravel(con.stat()), F_ref, 3)
+
+
+def test_low_level_fixed_effects():
+    # 
+    p = 100
+    X1, V1 = np.random.randn(p), np.ones(p)
+    Xf, Vf, tf = _fixed_effects([X1, 2 * X1], [V1, 4 * V1],
```
kchawla-pi: Fuller variable names will be helpful for someone like me.

kchawla-pi:
Add a whats_new entry.

examples/02_first_level_models/plot_fixed_effects.py
```diff
@@ -0,0 +1,112 @@
...
+#########################################################################
+# Statistics for the first session
+from nilearn import plotting
+cut_coords = [-129, -126, 49]
+
+fmri_glm = fmri_glm.fit(fmri_img[0], design_matrices=design_matrices[0])
+dic1 = fmri_glm.compute_contrast(contrast_val, output_type='all')
+plotting.plot_stat_map(
+    dic1['z_score'], bg_img=mean_img_, threshold=3.0, cut_coords=cut_coords,
+    title='%s, first session' % contrast_id)
```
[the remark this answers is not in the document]
jeromedockes: note that the python documentation explicitly states that there is no plan to deprecate printf-style formatting

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+    if len(contrast_imgs) != len(variance_imgs):
+        raise ValueError(
+            'The number of contrast images (%d) differs from the number of '
+            'variance images (%d). ' % (len(contrast_imgs), len(variance_imgs)))
+
+    # instantiate a masker
+    if isinstance(mask, NiftiMasker):
+        masker = mask
+    else:
+        masker = NiftiMasker(mask_img=mask).fit()
```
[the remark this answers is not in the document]
jeromedockes: if the user passes a NiftiMasker, we use it as it is (it must be fitted of course). if the user passes an image, we use it to initialize a masker

kchawla-pi:
I know. For new code, I would rather we use format.

examples/02_first_level_models/plot_fixed_effects.py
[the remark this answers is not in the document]
kchawla-pi: I know. For new code, I would rather we use .format. instead of printf style formatting.

examples/02_first_level_models/plot_fixed_effects.py
```diff
@@ -0,0 +1,112 @@
...
+from nistats.first_level_model import FirstLevelModel
+fmri_glm = FirstLevelModel(mask_img=data['mask'], smoothing_fwhm=5,
+                           minimize_memory=True)
+
+#########################################################################
+# Compute fixed effects of the two runs and compute related images
+# For this, we first define the contrasts as we would do for a single session
+n_columns = design_matrices[0].shape[1]
+
+def pad_vector(contrast_, n_columns):
```
jeromedockes: this is used only once in the example, so maybe it can be inlined and avoid the function definition

kchawla-pi: I like using the function, I am wondering if we should include this as an automatic feature at some point in the future because many examples seem to need padding.

examples/02_first_level_models/plot_fixed_effects.py
```diff
@@ -0,0 +1,112 @@
...
+difference is that the fixed-effects model is run explicitly here,
+after GLM fitting on two sessions.
+
+"""
+
+###############################################################################
+# Create a write directory to work
+# it will be a 'results' subdirectory of the current directory.
+from os import mkdir, path, getcwd
+write_dir = path.join(getcwd(), 'results')
```
jeromedockes: where is this used?

nistats/tests/test_first_level_model.py
```diff
@@ -59,6 +59,60 @@ def test_high_level_glm_one_session():
...
+             fmri_data,
+             mask,
+             multi_session_model,
+             rk,
+             shapes,
+             variance,
+             contrasts,
+             ffx_contrast, ffx_variance, ffx_stat,
+             dic1, dic2, ffx_dic,
+         )
```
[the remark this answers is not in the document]
bthirion: Sorry ?

nistats/tests/test_first_level_model.py
```diff
@@ -59,8 +59,8 @@ def test_high_level_glm_one_session():
     assert_true(isinstance(z1, Nifti1Image))
 
 
-def test_fixed_effects():
-    # New API
+def test_explicit_fixed_effects():
+    """ tests the fixed effects performed manyally/explicitly"""
```
kchawla-pi: manually

examples/02_first_level_models/plot_fixed_effects.py
```diff
@@ -0,0 +1,112 @@
...
+
+For details on the data, please see:
+
+Dehaene-Lambertz G, Dehaene S, Anton JL, Campagne A, Ciuciu P, Dehaene
+G, Denghien I, Jobert A, LeBihan D, Sigman M, Pallier C, Poline
+JB. Functional segregation of cortical language areas by sentence
+repetition. Hum Brain Mapp. 2006: 27:360--371.
+http://www.pubmedcentral.nih.gov/articlerender.fcgi?artid=2653076#R11
+
+Please see `plot_fiac_analysis.py` example for details.  The main
```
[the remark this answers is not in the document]
bthirion: Sorry, but how do I relate to another example ?

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+                         precision_weighted=False):
+    """Compute the fixed effets given images of effects and variance
+
+    Parameters
+    ----------
+    contrast_imgs: list of Nifti1Images or strings
+              the input contrast images
+    variance_imgs: list of Nifti1Images or strings
+              the input variance images
+    mask: Nifti1Image or NiftiMasker instance,
```
adelavega: It would be nice if this were an optional argument, and if a mask is not provided, that the mask can be learned like in `SecondLevelModel`. 

Otherwise, if the function that does the mask learning in `SecondLevelModel` can be abstracted to a more general function, that would make it easier to use that function prior to this one.

nistats/contrasts.py
```diff
@@ -276,3 +276,71 @@ def __rmul__(self, scalar):
...
+              Whether the fixed effects estimates should be weighted by inverse
+              variance or not. Defaults to False.
+
+    Returns
+    -------
+    ffx_contrast_img: Nifti1Image,
+             the fixed effects contrast computed within the mask
+    ffx_variance_img: Nifti1Image,
+             the fixed effects variance computed within the mask
+    ffx_t_img: Nifti1Image,
```
adelavega: Can other stats be returned like z? or p?

nistats/contrasts.py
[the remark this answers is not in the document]
adelavega: See above but it would be nice if you could learn a mask from the effect images as well.

bthirion:
OK, I'll take a look at your requests (that sound perfectly reasonable) this week.
Best,
B

nistats/contrasts.py
bthirion: Actually, there is a reason why I didn't do it: namely that you would need to provide (additionally) the degrees of freedom for each input, since the dofs of the output is basically the sum of the dofs of inputs. I was not sure it is worth the effort, given that the number of dofs is high in such models, so that eventually t~=z.  WDYT ?

adelavega: Ah, then I agree it's probably not worth it unless someone requests it. I don't personally need it.

adelavega:
Thanks. I'm going to give it a shot to incorporate this into fitlins and let you know if I run into any issues.

nistats/contrasts.py
```diff
@@ -312,6 +312,8 @@ def fixed_effects_img(contrast_imgs, variance_imgs, mask,
     # instantiate a masker
     if isinstance(mask, NiftiMasker):
         masker = mask
+    elif mask is None:
+        masker = NiftiMasker().fit(contrast_imgs)
```
adelavega: Would it be possible to add `smoothing_fwhm` as an optional kwarg? Some users may want to smoothing prior to this step.

It would also be similar to `SecondLevelModel` and `non_parametric_inference`

Maybe can copy this logic to apply it to both learned and passed in masks:
https://github.com/nistats/nistats/blob/f7e1056e26bd5b1070fc361a1d4335b14f0de32c/nistats/second_level_model.py#L635

adelavega: Also, a question. is doing `masker.fit(contrast_imgs)` equivalent to computing a mean img and fitting the mask to that (as is done in the SecondLevelModel)

adelavega: e.g.: 
```
    if isinstance(mask, NiftiMasker):
        masker = mask
    elif mask is None:
        masker = NiftiMasker(smoothing_fwhm=smoothing_fwhm).fit(contrast_imgs)
    else:
        masker = NiftiMasker(mask_img=mask, smoothing_fwhm=smoothing_fwhm).fit()
```

Minor comment: it seems other similar functions (such as `SecondLevelModel`, will call `.fit` even when a masker is passed, so this could confuse people)

bthirion: > Would it be possible to add `smoothing_fwhm` as an optional kwarg? Some users may want to smoothing prior to this step.

The problem is: what do you do with the variance ? you smooth it by the same amount ?
This is probably suboptimal (the variance obtained from the smoothed data would be smaller than the smoothed variance) and it makes it unclear what the statistic tests. 
So I'd rather heavily discourage it.

adelavega: Okay, I see. I suppose users can always pass a mask with their settings of choice if they want to do something else.

adelavega:
Are there plans to merge this, or will this be waiting until `nistats` gets merged into `nilearn`?

bthirion:
Sorry. I'll merge it tomorrow unless i hear otherwise.

bthirion:
Actually it requires a bit of work...

bthirion:
@kchawla-pi seems that I addressed your concerns. Can you confirm ?

kchawla-pi:
There are still some things. I will do them for you tomorrow and push to this branch.

bthirion:
Thx for the fixes and renaming. LGTM.

kchawla-pi:
@bthirion I'm merging this. If you think https://github.com/nistats/nistats/pull/386/files#r323429462 has merit then you can address it in a new PR.

examples/02_first_level_models/plot_fixed_effects.py
```diff
@@ -0,0 +1,112 @@
...
+difference is that the fixed-effects model is run explicitly here,
+after GLM fitting on two sessions.
```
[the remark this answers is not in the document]
bthirion: The point is: this is just an example showing that fixed effects can be done that way. 
The underlying use case would be that the user obtains summary statistics from a first level analysis and wants to obtain fixed effects statistics from these images. 
However, putting it that way does not really enlighten example... so I don't really see what to put.

PR #387, opened 2019-09-10
[MRG] HTML Scraper for GLM Reports

kchawla-pi:
- Added Elizabeth's version of HTML Scraper from Nilearn for GLM Reports


emdupre:
You have a hard-coded width of 1600px, which is pushing the "center" to the far right of the page ! You might want to add in some additional CSS specifically for sphinx reports -- here's what @GaelVaroquaux added :

```
/* Sphinx-gallery Report embedding */
div.sg-report {
    padding: 0pt;
    transform: scale(.95);
}

div.sg-report iframe {
    display: block;
    border-style: none;
    transform: scale(.85);
    height: 470px;
    margin-left: -12%; /* Negative because of .8 scaling */
    margin-top: -4%;
    padding: 0pt;
    margin-bottom: 0pt;
    width: 126%; /* More than 100% because of .8 scaling */
}
```

Probably the height will need to be adapted, since your reports are much longer than e.g. the NiftiMasker report. At a minimum, if I change your existing width to 100% it renders correctly !

kchawla-pi:
I am open to removing the hardcoding, if that means no specific for scraper CSS. Should I set it to 100% in the function cll?

GaelVaroquaux:
> I am open to removing the hardcoding, if that means no specific for scraper CSS.

Hard-coding width in px is dangerous indeed. I would suggest to set a
max-width, and set it in ex rather than in px.

kchawla-pi:
I had set the size to 1600px for the iframe so there wouldn't be horizontal scrolling in Jupyter notebooks, but it seems the width setting is not just for the iframe but also the HTML content within.

Removing the 1600px for the iframe gives this:

image

Will try to set the CSS tomorrow.

Issue #345, opened 2019-04-19
Remove the dependency on boto


GaelVaroquaux:
The dependency on boto is a blocker for integrating in nilearn. It should be removed from nistats.

It should be possible to implement the corresponding functionality (data download) with the Python standard library.

kchawla-pi:
I am looking into this now. Boto creates these urls to download from using some base path and I am guessing crawls through the server, finding every file that matches the filter to create a list of URLs to download from. The url list is  411759 entries.
I don't know where to begin to generate so many URLs using standar python.

effigies:
@kchawla-pi Would it make sense to have it as an optional dependency? You could add it somewhere in `extras_require` for `setuptools` and then have any internal logic use [nibabel.optpkg](https://github.com/nipy/nibabel/blob/master/nibabel/optpkg.py):

```Python
from nibabel.optpkg import optional_package
boto3, has_boto3, setup_module = optional_package('boto3')
```

If it's missing, attempts to use the `boto3` "module" will result in an informative error asking users to install it. And you can condition some things on `has_boto3`.

kchawla-pi:
It's already optional, only imported when a specific dataset fetcher is called. Currently it raises an error, instead of asking nicely, but that is easily fixed.

GaelVaroquaux:
But that means that examples cannot run without boto, which makes
teaching harder.

I don't understand why it should be mandatory. There must be a way to
query the server to get the file list.

jeromedockes:
> The dependency on boto is a blocker for integrating in nilearn. It should be removed from nistats.
> 

 +1

> It should be possible to implement the corresponding functionality (data download) with the Python standard library.

yes. although at some point we could consider adding `requests` as a dependency, since it is ubiquitous and much easier to use than the standard library modules

GaelVaroquaux:
> although at some point we could consider adding requests as a
> dependency, since it is ubiquitous and much easier to use than the
> standard library modules

Yes.

kchawla-pi:
>  although at some point we could consider adding requests as a
>  dependency, since it is ubiquitous and much easier to use than the
>  standard library modules

> Yes.

+1
Then we should slowly and systematically replace existing solutions with requests, else we'll be trading one dependency for another if we only use it once.

kchawla-pi:
A workaround for this issue can be that we upload the urls (currrently saved as a json file by the fetcher to osf.io and the fetcher downloads that, and uses that information to download the dataset itself.
It should not require much work.

[...]

PR #393, opened 2019-09-18
[MRG] Remove Boto3 dependency

kchawla-pi:
Fixes issue #345 

Removed Boto3 need by directly downloading `urls.json` file we uploaded to osf.io
As posited in https://github.com/nistats/nistats/issues/345#issuecomment-532654217


bthirion:
But you haven't removed the dependence on boto yet ?

kchawla-pi:
@bthirion Ready for review.

nistats/_utils/datasets.py
```diff
@@ -0,0 +1,93 @@
...
+    since this approach is faster and circumvents `boto` as a dependency.
+
+    If the URLs list become outdated in future, this function
+    creates a file 'nistats_fetcher_openneuro_dataset_urls.json'
+    with an updated list of URLs, ready to be uploaded as quick-files
+    to Nilearn account on osf.io.
+    To ensure that the same download link is updated,
+    the file name must remain the same.
+
+    Note: This function requires ``boto3`` to be installed.
```
bthirion: There is no way to avoid this ?

.travis.yml
```diff
@@ -44,7 +44,7 @@ matrix:
     - env: DISTRIB="conda" PYTHON_VERSION="*"
            NUMPY_VERSION="*" SCIPY_VERSION="*"
            SCIKIT_LEARN_VERSION="*" MATPLOTLIB_VERSION="*" COVERAGE="true"
-           PANDAS_VERSION="*"
+           PANDAS_VERSION="*" BOTO3=true
```
bthirion: By default, BOTO3='' ?

nistats/datasets.py
```diff
@@ -82,14 +64,18 @@ def fetch_bids_langloc_dataset(data_dir=None, verbose=1):
...
+                                  verbose=1):
+    """ Download a file with OpenNeuro BIDS dataset index.
 
     Downloading the index allows to explore the dataset directories
     to select specific files to download. The index is a sorted list of urls.
 
-    Note: This function requires boto3 to be installed.
+    Notes
+    -----
+    Update this list (usually not necessary, requires `boto3`) using
```
bthirion: But does one know that this is necessary ?

nistats/datasets.py
```diff
@@ -241,8 +209,13 @@ def fetch_openneuro_dataset(
 
     downloaded_files: list of string
         Absolute paths of downloaded files on disk
+
+    Notes
+    -----
+    If necessary (rarely), outdated dataset index can be updated using
+    ``nistats._utils.datasets.make_openneuro_dataset_urls_index``
+    (requires `boto3`).
```
bthirion: How will the user diagnose that it is necessary ?

.travis.yml
kchawla-pi: This is the only build where BOTO3 dependent function is tested. There doesn't need to be a BOTO3 env var in other builds.

kchawla-pi:
@bthirion Rewrote documentation for clarity.

nistats/_utils/datasets.py
kchawla-pi: No. This function updates the URL index we have uploaded to osf.io, in case OpenNeuro changes the layout .
Boto3 is used in this function (not using it is too much work for almost no benefit). But it is no longer a user facing dependency.

kchawla-pi:
@bthirion this needs review

PR #400, opened 2019-10-09
[MRG] remove file local variables

jeromedockes:
some nistats set file local variables for emacs and vim. it is redundant because
it sets them to the defaults of any reasonable configuration, potentially
annoying if these variables have not been marked as safe, and it is not
efficient as it is not done systematically, nor in any nilearn files, nor for
other editors. I therefore suggest we remove these lines before merging into
nilearn.

from nistats dir:
```
   find -name '*.py' -exec sed -i '/^# emacs: -\*-.*/d' {} \;
   find -name '*.py' -exec sed -i '/^# vi: set.*/d' {} \;
```


bthirion:
I'll merge tomorrow unless I hear otherwise.

PR #403, opened 2019-10-15
Use a get data function

jeromedockes:
`get_data` is deprecated in nibabel

### DIFF
--- a/nistats/reporting/_get_clusters_table.py
+++ b/nistats/reporting/_get_clusters_table.py
@@ -116,7 +116,7 @@ def get_clusters_table(stat_img, stat_threshold, cluster_threshold=None,
         reports the center of mass of the cluster, rather than any peaks/subpeaks.
     """
     cols = ['Cluster ID', 'X', 'Y', 'Z', 'Peak Stat', 'Cluster Size (mm3)']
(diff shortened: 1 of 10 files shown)


nistats/reporting/_get_clusters_table.py
```diff
@@ -12,6 +12,7 @@
 import nibabel as nib
 from scipy import ndimage
 from nilearn.image.resampling import coord_transform
+from nistats.utils import get_data
```
bthirion: I'd rather use it from nilearn directly rather than copying to nistats.

jeromedockes:
> I'd rather use it from nilearn directly rather than copying to nistats.

but that would fail for users who don't have a recent version of nilearn

kchawla-pi:
>> I'd rather use it from nilearn directly rather than copying to nistats.

> but that would fail for users who don't have a recent version of nilearn

Jerome is right. If we are to stop Nistats development and keep it for users in as complete state as possible it is best to incorporate this whole feature. 
Other wise, we might as well not bother since Nistats won't be long once we release it as nilearn.stats

bthirion:
Merging --- to be on par with Nilearn.
Thx !

PR #408, opened 2019-11-07
Reduce memory by deleting unmasked run_img

adelavega:
The `run_img` object is not needed after its masked. Deleting it saves (in my case) around 10x the size of the compressed nifti, which frees up a lot of memory for model fitting.

To give more background, I'm trying to run `nistats` on a cloud server with limited memory. It looks like the bottleneck is not model fitting (peaks at ~2 GB with 2 predictors), but rather loading the unmasked nifti. What is a bit confusing to me is that in this instance I have not passed a mask, so `nilearn` is using a dummy mask of all 1s. Yet, the memory foot print of `Y` is about 6-7x smaller than `run_img`. 

Is there any way to load `run_img` in a way that uses less memory to being with?

In any case, the deletion in this PR helps out a bit.

### DIFF
--- a/nistats/first_level_model.py
+++ b/nistats/first_level_model.py
@@ -463,7 +463,7 @@ class FirstLevelModel(BaseEstimator, TransformerMixin, CacheMixin):
                 sys.stderr.write('Starting masker computation \r')
 
             Y = self.masker_.transform(run_img)
-            del run_img
+            del run_img  # Delete unmasked image to save memory
 
             if self.verbose > 1:
                 t_masking = time.time() - t_masking
--- a/nistats/first_level_model.py
+++ b/nistats/first_level_model.py
@@ -8,7 +8,6 @@ of fMRI data analyses.
 Author: Bertrand Thirion, Martin Perez-Guevara, 2016
 
 """
-
 import glob
 import json
 import os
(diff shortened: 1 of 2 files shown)


adelavega:
The discrepancy I observed was due to the fact that the image object was being recast to `float64` and then back down to `float32`. The masked object was in `float32` so deleting the `float64` `run_img` object helps keep the memory usage (in my test case) to 3.9 GB versus 5.5GB originally. 

Loading and recasting the `run_img` prior to `model.fit` helped keep my memory usage even lower at ~3.5GB. 

The casting to `float64` seems to be a bug (to be filed soon) upstream in nibabel. 

Either way, I still suggest merging this PR as that `run_img` is potentially still using memory unnecessarily.

bthirion:
LGTM. Will merge tomorrow unless i hear otherwise.

adelavega:
Heres the nibabel bug: https://github.com/nipy/nibabel/issues/832. Sounds good.

kchawla-pi:
We need this in changes section in whats new.

nistats/first_level_model.py
```diff
@@ -464,6 +463,7 @@ def fit(self, run_imgs, events=None, confounds=None,
                 sys.stderr.write('Starting masker computation \r')
 
             Y = self.masker_.transform(run_img)
+            del run_img
```
kchawla-pi: Add a comment (your audience is someone new to the project) why this is being done.

adelavega: Sure, although I see that deleting unused variables happens throughout the code base (for example in `run_glm`). So this doesn't seem any different from any other instance of memory saving practices throughout nistats.

bthirion:
Thx, merging.

PR #402, opened 2019-10-15
[WIP] All-Resolution Inference

bthirion:
Introduction of a cluster-level inference procedure based on the all-resolution inference procedure.

PR #413, opened 2019-12-08
Undo accidental overwriting of the @bthirion  PR #402

kchawla-pi:
This is to fix the mess I made with bthirion's PR #402


kchawla-pi:
Bertrand used his local commit history to open a new PR. so this is obviated.

PR #415, opened 2019-12-18
Fix: Missing variables in localizerd dataset are no longer None, but n/a

bthirion:
missing variables in localizerd dataset are no longer None, but n/a


kchawla-pi:
Ha! I noticed this yesterday when diagnosing CircleCI. Glad you took care of it. Thanks!

Issue #409, opened 2019-11-08
Misc issues in Nistats reports


bthirion:
Here are some  issues I have encountered using nistats reports.
* Does not seem to take into account use-provided parameters for thresholding 
* the contrast maps only have the positive part displayed
* multi-dimensional contrasts do not display well.

kchawla-pi:
> the contrast maps only have the positive part displayed

That's due to issue #362 

> Does not seem to take into account use-provided parameters for thresholding
> multi-dimensional contrasts do not display well.

Examples/code to reproduce?

bthirion:
Some of my comments are wrong. Let me correct them:
- there is a slight inconsistency in the default parameters. default alpha should 0.001 not 0.01, to match the default threshold (3.09)
- #362 
- the display of multi-dimensional contrasts is not good.

Clearly these 3 things are distinct and thus should be handled in separate PRs. I Can work on it.

danjgale:
I've also encountered an issue with the reports. If the inputs to your second-level analysis are a list of  `FirstLevelModel` instances, there is no way to define first-level contrasts of these models. In this case, defining first-level contrasts are necessary, otherwise `SecondLevelModel.compute_contrasts` will throw an error: 

```
ValueError: If second_level_input was a list of FirstLevelModel, then first_level_contrast is mandatory. It corresponds to the second_level_contrast argument of the compute_contrast method of FirstLevelModel
```. 

It seems to me that if you pass in a second-level model, it only accepts second-level contrasts. Not sure whether or not to make a separate issue with the impending nilearn merge.

bthirion:
Thx for reporting.
I think that this is a sepaarte issue, that should be addressed any how.

kchawla-pi:
> Not sure whether or not to make a separate issue with the impending nilearn merge.

I've made a separate issue. It can be worked on in `nilearn.stats` after the merge.

danjgale:
Thanks @kchawla-pi, I'd be happy to work on it once the merge is done!

PR #418, opened 2019-12-30
[WIP] Improve reporting: Correct alpha default value & enable 2 sided map_threshold

bthirion:
Addressing #409


nistats/thresholding.py
```diff
@@ -88,20 +94,26 @@ def map_threshold(stat_img=None, mask_img=None, alpha=.001, threshold=3.,
...
     if height_control not in ['fpr', 'fdr', 'bonferroni', None]:
         raise ValueError(
-            "height control should be one of ['fpr', 'fdr', 'bonferroni', None]")
+            "height control should be one of "
+            "['fpr', 'fdr', 'bonferroni', None]")
 
+    # if two-sided, correct alpha by a factor of 2
+    alpha_ = alpha
+    if two_sided:
+        alpha_ = alpha / 2
```
kchawla-pi:
```python
alpha_corrected = alpha / 2 if two_sided else alpha
```

nistats/thresholding.py
```diff
@@ -88,20 +94,26 @@ def map_threshold(stat_img=None, mask_img=None, alpha=.001, threshold=3.,
...
+    
     # if height_control is 'fpr' or None, we don't need to look at the data
     # to compute the threhsold
     if height_control == 'fpr':
-        threshold = norm.isf(alpha)
+        threshold = norm.isf(alpha_)
```
kchawla-pi: ~`alpha_`~  `alpha_corrected`

kchawla-pi: So on in the rest of the code.

kchawla-pi:
@bthirion I plan to implement my suggestions and merge this one soon.

nistats/thresholding.py
bthirion: No, we should not name it 'alpha_corrected', which alludes to something else in this context.
To me the trailing _ simply means that the value had to be adapted etc. Why is it bad ?

kchawla-pi: It would be clearer to have a more explicit and self-documenting name. If not `alpha_corrected` then perhaps `alpha_adapted`?

kchawla-pi:
Any opinions on renaming `alpha_`?
Also, add a whats_new entry, after pulling from master.

PR #424, opened 2020-01-10
GLM Reports: Computed masks are used if there is no user supplied one

kchawla-pi:


nistats/reporting/glm_reporter.py
```diff
@@ -202,7 +202,8 @@ def make_glm_report(model,
                                                    )
     statistical_maps = make_stat_maps(model, contrasts)
     html_design_matrices = _dmtx_to_svg_url(design_matrices)
-    mask_plot_html_code = _mask_to_svg(mask_img=model.mask_img,
+    mask_img = model.mask_img or model.masker_.mask_img
```
bthirion: Sorry, I'm just surprised by this syntax... 
If only one of the two exists, I understand the result, but what happens if None of the two mask_img is defined, or if both masks exist and are different ?

kchawla-pi: If `model.mask_img` eavluates to true, that's what gets assigned, if it evaluates to false, `model.masker_.mask_img` gets assigned, irrespective of the value of `model.masker_.mask_img`.
I will add an explicit `None` option as well. 

Is it possible for the `model.masker_.mask_img` to not exist?

bthirion: I think that sometimes, only `model.masker_.mask_img_` is defined.

kchawla-pi: `model.masker_.mask_img_` is always defined then that's fine. The current code will prefer `model.mask_img` but if that is falsey, witll us `model.masker_.mask_img_`.
My question is specifically if there's a y case where `model.masker_.mask_img_` does not exist. Do I need to account for a `NameError`?

bthirion: if model.masker_ has been fitted, then IIRC the corresponding mask_img_ is defined.

PR #427, opened 2020-01-13
Removed the Sphinx Scraper for HTML (new Sphinx Gallery version has it)

kchawla-pi:
Same as https://github.com/nilearn/nilearn/pull/2261


kchawla-pi:
The report is not being inserted. Will have to look into this.
Or maybe close this PR, and fix it in Nilearn instead.

kchawla-pi:
What's working for Nilearn isn't working here. I would rather debug this in `nilearn.stats` I am closing this, and removing the `report` variable insertions to prevent the duplicates for the nistats repo.

PR #430, opened 2020-01-16
Fixing Flake8 problems

kchawla-pi:
Before merging in Nilearn, fixing all the flake 8 inconsistencies here


nistats/_utils/testing.py
```diff
@@ -124,14 +124,17 @@ def _create_fake_bids_dataset(base_dir='', n_sub=10, n_ses=2,
             func_path = os.path.join(subses_dir, 'func')
             os.makedirs(func_path)
             for task, n_run in zip(tasks, n_runs):
-                for run in ['run-%02d' % label for label in range(1, n_run + 1)]:
+                for run in ['run-%02d' % label
+                            for label in range(1, n_run + 1)
+                            ]:
```
bthirion: Is this really required ? Personally I don't like this (yet I agree that style is not a matter of taste)

nistats/hemodynamic_models.py
```diff
@@ -513,12 +517,13 @@ def compute_regressor(exp_condition, hrf_model, frame_times, con_id='cond',
     # 4. temporally resample the regressors
     if hrf_model == 'fir' and oversampling > 1:
         computed_regressors = _resample_regressor(
-            conv_reg[:, oversampling - 1:], hr_frame_times[: 1 - oversampling],
-            frame_times)
+                conv_reg[:, oversampling - 1:],
```
bthirion: 8-space indent ?

nistats/hemodynamic_models.py
```diff
@@ -513,12 +517,13 @@ def compute_regressor(exp_condition, hrf_model, frame_times, con_id='cond',
...
+                hr_frame_times[: 1 - oversampling],
+                frame_times)
     else:
         computed_regressors = _resample_regressor(
-            conv_reg, hr_frame_times, frame_times)
-        
+                conv_reg, hr_frame_times, frame_times)
```
bthirion: Same here ?

nistats/hemodynamic_models.py
kchawla-pi: So long as it's a multiple of 4, this is pretty common.

nistats/hemodynamic_models.py
kchawla-pi: idem.

bthirion:
I'm sorry, but we have always used 4. We need to stick to that.

kchawla-pi:
It's not an 8 space indent, it is double indented (2 indents of 4 spaces each), what is called a continuation indent, automatically used when we bring the first parameter or arguments to the next line after the opening parenthesis.
If it was a different type of indentation, Flake8 would have red-flagged it.
By default, many IDEs make the continuation indent double indented to make it easier to see that the code block is not different, it is the same code block on a new line.

kchawla-pi:
Continuation or hanging indent.
This should clarify this, from PEP8:
https://www.python.org/dev/peps/pep-0008/#indentation

nistats/hemodynamic_models.py
bthirion: Thx for the pointer, but _sorry to insist on such sow-level things_ I don't see the rationale here, since there is no possible confusion with nested indents.
Unfortunately, this pattern is present in lost of places in this PR. 
I'm usually not very picky on style considerations, but in view of the integration into Nielarn, that enforces this patterns, I really consider we should stick to the 4-space indents whenever this is not problematic.

PR #435, opened 2020-01-30
[WIP] Sphinx should only build files that have changed

kchawla-pi:


bthirion:
Is that one still alive ?
Best,

